PR #1, opened 2016-08-21
Update plugin-build-metrics.yml

opx3:
I would like to be able to release my plugin


daniel-beck:
Reference: https://github.com/jenkinsci/build-metrics-plugin/commits/master

PR #2, opened 2016-08-22
Required plugin for blue ocean

michaelneale:
cc @daniel-beck @vivek


daniel-beck:
@michaelneale Looks good.

michaelneale:
@daniel-beck btw this is for this plugin: https://github.com/jenkinsci/blueocean-plugin/tree/master/blueocean-config so I trust it follows the correct pattern.

PR #6, opened 2016-08-23
Add new plugin: microdocs-integration

steven166:


daniel-beck:
@steven166 Won't the Maven coordinates be `com.maxxton.microdocs-integration`?

https://github.com/jenkinsci/microdocs-integration-plugin/blob/master/build.gradle#L1

steven166:
@daniel-beck Ah yes, I'll change it

PR #8, opened 2016-08-24
Added 'praqma' (jenkins-wiki user) to access list

MadsNielsen:
I've added the user 'praqma' to both ClearCaseUCM plugin and ConfigRotator plugin. This user does not have a github account, but he has upload permissions.


MadsNielsen:
It should be noted that 'praqma' is an automation user, that is only used when we release components in our pipelines. So he's not exactly able to comment in this context :)

batmat:
:+1:

oleg-nenashev:
👍

PR #10, opened 2016-08-25
Add snapshot per-version maven-metadata.xml to list

daniel-beck:


oleg-nenashev:
👍

batmat:
:+1: 

Seems correct. Manually picked up a few plugins under http://repo.jenkins-ci.org/public/org/jenkins-ci/plugins/ to see if I didn't catch anything obvious.

In the go, I found https://support.sonatype.com/hc/en-us/articles/213464098-What-are-the-requests-that-Maven-3-x-sends-when-deploying-artifacts- which is an interesting reference document about the requests being sent, and hence the requests/paths to allow.

batmat:
I'm trying to see if there's some new hash format likely to appear (other than md5 or sha1, I mean). Does not seem to.

PR #11, opened 2016-08-25
Remove obsolete paths as workaround for INFRA-905

daniel-beck:
Artifactory appears to have a limit of probably (undocumented) 1024 chars (Office-365-Connector had 1026 chars in its `includePattern`, the other affected ones were all longer than that).

Keeps the most recently used path/groupId for a given artifact.

To be merged only if JFrog support has no better solution. Or maybe even then. Do we want to prune obsolete paths?

### DIFF
--- a/permissions/component-deployment-sphere.yml
+++ b/permissions/component-deployment-sphere.yml
@@ -1,8 +1,6 @@
 ---
 name: "deployment-sphere"
 paths:
-- "com/epam/jenkins/deployment-sphere/deployment-sphere"
-- "org/jenkins/plugins/deployment-sphere"
 - "org/jenkins-ci/plugins/deployment-sphere"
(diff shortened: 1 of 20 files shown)


orrc:
Removing old paths makes things more secure, rather than less secure, so sounds reasonable to me, if JFrog don't have a solution.

daniel-beck:
Shouldn't really matter all that much from a security POV, and may prevent something like maintenance branch releases using the 'obsolete' pattern.

Extremely unlikely, admittedly, but who knows…

batmat:
:+1: good idea to avoid mistakes. It's so easy to fix if necessary that it's worth doing indeed.

daniel-beck:
JFrog support was unable to offer a workaround.

PR #12, opened 2016-08-25
Add upload permissions for hyper-build-step-plugin

Jimmy-Xu:


daniel-beck:
Corresponding repository / POM:
https://github.com/jenkinsci/hyper-build-step-plugin/blob/master/pom.xml#L11...L12

@Jimmy-Xu is one of the maintainers and has commit access to that repo.

daniel-beck:
@Jimmy-Xu Permissions should be active within 5 minutes.

PR #16, opened 2016-08-28
Adding abayer to branch-api.

abayer:
Need to deploy a SNAPSHOT for dependencies.

cc @daniel-beck @oleg-nenashev


daniel-beck:
Since you're already a committer, I will just merge and notify @jglick @stephenc @recena

PR #17, opened 2016-08-29
Adding new plugin pipeline-model-definition

abayer:
cc @oleg-nenashev @daniel-beck

### DIFF
--- a/permissions/plugin-pipeline-config.yml
+++ b/permissions/plugin-pipeline-config.yml
@@ -1,7 +1,7 @@
 ---
-name: "pipeline-config"
+name: "pipeline-model-definition"
 paths:
-- "org/jenkinsci/plugins/pipeline-config"
+- "org/jenkinsci/plugins/pipeline-model-definition"
 developers:
 - "abayer"
 - "oleg_nenashev"
--- a/permissions/plugin-pipeline-config.yml
+++ b/permissions/plugin-pipeline-config.yml
@@ -1,0 +1,7 @@
+---
+name: "pipeline-config"
+paths:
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
👎 since @abayer didn't respond to the comment in the hosting request

abayer:
One sec...missed that.

permissions/plugin-pipeline-config.yml
```diff
@@ -0,0 +1,7 @@
+---
+name: "pipeline-config"
+paths:
+- "org/jenkinsci/plugins/pipeline-config"
+developers:
+- "abayer"
+- "oleg-nenashev"
```
daniel-beck: Wrong user name.

batmat:
Well, I guess this one is to be merged by @oleg-nenashev who should know quite better :-)

permissions/plugin-pipeline-config.yml
```diff
@@ -0,0 +1,7 @@
+---
+name: "pipeline-model-definition"
+paths:
+- "org/jenkinsci/plugins/pipeline-model-definition"
```
oleg-nenashev: Are you sure there is no dash?

abayer: Yup. My instincts are always to avoid the dash in groupId. I can change it if you'd prefer.

oleg-nenashev: no strong opinion. We have no standard from this, and plugin devs use whatever they want

daniel-beck: Ideally should just be what's the default presented by `hpi:create` :-)

PR #19, opened 2016-08-31
Create plugin-display-url-api.yml

i386:
PTAL @oleg-nenashev


batmat:
Seems fine. I suppose you checked with the team this is indeed their Jenkinsci accounts id, and not GitHub :).

PR #20, opened 2016-08-31
Did not include artifactId in path

i386:
Did not include artifactId in path


daniel-beck:
@batmat FYI https://github.com/jenkins-infra/repository-permissions-updater/blob/master/doc/yml-example.png

PR #21, opened 2016-08-31
Remove ivan

i386:


daniel-beck:
For the record, `imeredith` apparently has never logged in to Artifactory, which is required for the (LDAP-backed) account to get created. Therefore Artifactory rejected this permission target with

> "Permission target contains a reference to a non-existing user 'imeredith'."

I will update the documentation accordingly, and will implement a retry mode that strips out non-existent user names.

PR #31, opened 2016-09-06
Added/Edit permissions for absint plugin

bschommer:
The developer for astree are jherter and absint. Also added
permissions for a3.

### DIFF
--- a/permissions/plugin-absint-a3.yml
+++ b/permissions/plugin-absint-a3.yml
@@ -1,0 +1,6 @@
+---
+name: "absint-a3"
+paths:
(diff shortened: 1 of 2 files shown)


daniel-beck:
@bschommer Has that account logged in at repo.jenkins-ci.org once? Please confirm.

Also, no point in merging while @jherter hasn't logged in yet -- the permission change won't be applied (an issue in the current implementation -- when the request fails because a user doesn't exist, the tool doesn't retry without that user).

jherter:
Both accounts have logged in now.

daniel-beck:
Permissions have been updated.

PR #35, opened 2016-09-10
add imod as commiter of plugin-managed-scripts

imod:
@daniel-beck

### DIFF
--- a/permissions/plugin-managed-scripts.yml
+++ b/permissions/plugin-managed-scripts.yml
@@ -3,4 +3,4 @@ name: "managed-scripts"
 paths:
 - "org/jenkinsci/plugins/managed-scripts"
 developers:
-- "domi"
+- "imod"


orrc:
Are you sure this is correct? The user IDs in these files should be your Jenkins account name, and not your GitHub username.

According to Artifactory, the last release of this plugin was made by `domi`:
https://repo.jenkins-ci.org/webapp/#/artifacts/browse/tree/General/releases/org/jenkinsci/plugins/managed-scripts/1.2.1

orrc:
If you're being rejected when uploading, you might need to log in to Artifactory first (and then potentially wait 30 mins until this tool runs again?):
https://github.com/jenkins-infra/repository-permissions-updater#requesting-permissions

imod:
I used to have two accounts "domi"/"imod" on on the jenkins infra - but to be honest, I have no idea. 
But did perform a release of the config-file-provider-plugin just yesterday and it worked with 'imod' in my settings.xml (although it has both of my accounts in the file in this repo).

imod:
...but the release with "domi" always fails - so I guess I should only use "imod". The only reason I had two accounts was because "domi" was not recoverable for some reasons at that time. Anyway "imod" it should be...

imod:
thanks, now it worked: https://repo.jenkins-ci.org/webapp/#/artifacts/browse/tree/General/releases/org/jenkinsci/plugins/managed-scripts/1.2.4

PR #36, opened 2016-09-11
add codebeamer-coverage-publisher-plugin

abanfi:
add balazstar to codebeamer-coverage-publisher-plugin


oleg-nenashev:
Merge conflict

oleg-nenashev:
I suspect somebody has already added the plugin

PR #44, opened 2016-09-14
Add two new modules/POMs to deploy for the TFS plugin

olivierdagenais:
I recently split the TFS plugin project into three pieces: the `tfs-parent` top-level project, the `tfs-sdk` module and the `tfs` module (the original project).

Is it sufficient to add these files or do the folders need to be created in Artifactory, first?

Thanks!
\- Oli


oleg-nenashev:
I suppose "pom-tfs-sdk.yml" should be actually renamed to "lib-tfs-sdk.yml"

olivierdagenais:
> I suppose "pom-tfs-sdk.yml" should be actually renamed to "lib-tfs-sdk.yml"

If you want, sure; I just didn't see any precedent for that convention and the documentation only called out `component`, `plugin`, or `pom` as prefixes.  Would `component-tfs-sdk.yml` fit better?

Let me know either way and I can make the change.

Thanks!
\- Oli

oleg-nenashev:
Maybe "component" then

olivierdagenais:
> Maybe "component" then

Done!

olivierdagenais:
Confirmed, thanks!

PR #47, opened 2016-09-19
Add impersonation plugin

stephenc:


daniel-beck:
Plugin repo: https://github.com/jenkinsci/impersonation-plugin

PR #51, opened 2016-09-22
add some more folks to permissions for releasing blue ocean

michaelneale:
cc @daniel-beck @rtyler

### DIFF
--- a/permissions/plugin-blueocean-analytics-tools.yml
+++ b/permissions/plugin-blueocean-analytics-tools.yml
@@ -5,3 +5,6 @@ paths:
 developers:
 - "michaelneale"
 - "vivek"
+- "tfennelly"
+- "kzantow"
+- "tscherler"
--- a/permissions/plugin-blueocean-commons.yml
+++ b/permissions/plugin-blueocean-commons.yml
@@ -5,3 +5,6 @@ paths:
 developers:
 - "michaelneale"
 - "vivek"
+- "tfennelly"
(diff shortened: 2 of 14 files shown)


daniel-beck:
What about https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/pom-blueocean-parent.yml ?

michaelneale:
*sigh* this is getting tedious. 

OK added those @daniel-beck  (didn't even know about that file)

michaelneale:
dat logo ;)

oleg-nenashev:
👍

daniel-beck:
@michaelneale 

> sigh, really wish didn't have to do this

As do I. If you have a better idea, I'd like to know. (Combining all related components into one file may work for this particular zoo of plugins, but little else.)

michaelneale:
@daniel-beck yeah its mostly tedious for multi module plugins (which have their own issues, and probably aren't always a good idea). Security is often an enemy in ease of use by necessity, sadly..

daniel-beck:
@michaelneale FYI at least @scherler has not logged in yet, resulting in permissions update failure.

https://github.com/jenkins-infra/repository-permissions-updater#requesting-permissions

michaelneale:
@daniel-beck yeah hopefully he will. Ideally this could skip over mistaken user id's.

daniel-beck:
@michaelneale Permissions are now getting applied.

PR #56, opened 2016-09-26
Add MadsNielsen and Praqma to post build task plugin

MadsNielsen:
Im adding this request to adopt the plugin "Hudson Post Build Task" plugin. 

I've added my user MadsNielsen and our automation user Praqma to commit acces.

### DIFF
--- a/permissions/plugin-postbuild-task.yml
+++ b/permissions/plugin-postbuild-task.yml
@@ -2,4 +2,6 @@
 name: "postbuild-task"
 paths:
 - "org/jvnet/hudson/plugins/postbuild-task"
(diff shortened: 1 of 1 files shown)


daniel-beck:
@MadsNielsen To clarify, did you message the developers' list (ideally CC the previous maintainer) about adopting the plugin, or otherwise communicated with the former maintainer? If not, that should go first. If so, could you point me to any reference for that?

daniel-beck:
Dev list thread reference: https://groups.google.com/forum/#!topic/jenkinsci-dev/0l0443e10hc

PR #65, opened 2016-10-03
created plugin-pipeline-aggregator-view.yml

paul8620:
I am adding the pipeline-aggregator-view plugin to the permissions.


daniel-beck:
Wrong groupId in `path`: https://github.com/jenkinsci/pipeline-aggregator-view-plugin/blob/master/pom.xml#L13

Please see the example in https://github.com/jenkins-infra/repository-permissions-updater#managing-permissions

paul8620:
Changed the path and the id to match the same, had an issue figuring that part out didn't actually knew what to put in path, now it's clear thanks

PR #66, opened 2016-10-04
add robkuehn to Performance Signature plugins as maintainer

rpionke:
robkuehn is a colleague of mine


daniel-beck:
@rpionke Please note that `robkuehn` needs to log in to Artifactory and wait for ~30 minutes before this change will be applied. See prerequisite at https://github.com/jenkins-infra/repository-permissions-updater#requesting-permissions

PR #73, opened 2016-10-07
Work around the artifactory length limitation for docker-slaves

daniel-beck:
Artifactory restricts the length of certain fields in permission targets (and JFrog rejected my bug), and the current version of docker-slaves hits that limit, therefore the permissions update gets rejected. We need to remove at least one of the paths to make this work again.

@ndeloof Does this work for you like this, or want to retain one of the old paths?

@orrc @batmat @oleg-nenashev

### DIFF
--- a/permissions/plugin-docker-slaves.yml
+++ b/permissions/plugin-docker-slaves.yml
@@ -1,8 +1,6 @@
 ---
 name: "docker-slaves"
 paths:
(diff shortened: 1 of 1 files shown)


ndeloof:
interesting bug ...

orrc:
👍

daniel-beck:
Holding off the merge because @ndeloof filed #74 and it's not clear why.

PR #74, opened 2016-10-07
Update plugin-docker-slaves.yml

ndeloof:
see https://github.com/jenkins-infra/repository-permissions-updater/pull/73


daniel-beck:
@ndeloof Unclear why this exists, could you clarify? It looks identical to #73…

PR #80, opened 2016-10-12
Adding new pipeline-model-api plugin and pipeline-model-parent.

abayer:
We're going to split pipeline-model-definition up into two plugins, so
we need to be able to push the new plugin and its parent.

### DIFF
--- a/permissions/plugin-pipeline-model-definition.yml
+++ b/permissions/plugin-pipeline-model-definition.yml
@@ -5,3 +5,4 @@ paths:
 developers:
 - "abayer"
 - "oleg_nenashev"
+- "stephenconnolly"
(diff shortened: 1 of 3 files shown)


daniel-beck:
This is related to https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/plugin-pipeline-config.yml right? Does it make sense to include a different set of users here?

Also, file name really shouldn't have been `plugin-pipeline-config.yml`, please rename to match contents if you're going to touch it to add stephenc.

abayer:
@daniel-beck Done and done.

permissions/plugin-pipeline-model-definition.yml
```diff
@@ -5,3 +5,4 @@ paths:
 developers:
 - "abayer"
 - "oleg_nenashev"
+- "stephenconnolly"
```
batmat: NIT: missing newline compared to other files

abayer:
Upcoming PR.

PR #84, opened 2016-10-14
Add myself as a developer

stephenc:


daniel-beck:
@jglick Delete is not granted to anyone.

PR #91, opened 2016-10-18
Add puppet plugin permissions

ccaum:
This adds the necessary permissions for the puppet-enterprise-pipeline
plugin.

### DIFF
--- a/permissions/plugin-puppet-enterprise.yml
+++ b/permissions/plugin-puppet-enterprise.yml
@@ -1,6 +1,6 @@
 ---
 name: "puppet-enterprise-pipeline"
 paths:
-- "org/puppet/plugins/puppet-enterprise-pipeline"
+- "org/jenkins-ci/plugins/workflow/puppet-enterprise-pipeline"
 developers:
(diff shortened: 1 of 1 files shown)


permissions/plugin-puppet-enterprise.yml
```diff
@@ -0,0 +1,6 @@
+---
+name: "puppet-enterprise-pipeline"
+paths:
+- "org/puppet/plugins/puppet-enterprise-pipeline"
```
daniel-beck: According to the pom.xml, your groupId is `org.jenkins-ci.plugins.workflow`, so this should be `org/jenkins-ci/plugins/workflow/puppet-enterprise-pipeline`.

daniel-beck:
Repo: https://github.com/jenkinsci/puppet-enterprise-pipeline-plugin

permissions/plugin-puppet-enterprise.yml
ccaum: Just pushed e55221e to fix that. Thanks

PR #96, opened 2016-10-24
Adding new pipeline-model-declarative-agent plugin.

abayer:
This is also in the pipeline-model-definition-plugin repo, currently
in a PR that should be getting merged shortly, so I'm being a bit
proactive.

Also added Bobby as someone who can release the suite.


daniel-beck:
Reference: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/38

PR #104, opened 2016-11-02
Add praqma user as developer, our automation user

MadsNielsen:
We'll need to add our company automation user "praqma" as a commiter here. He doesn't really log into github so can we add him?


daniel-beck:
No problem, as you're the maintainer, you can add whoever you like :+1:

PR #111, opened 2016-11-06
add permission for thinbackup and SCTMExecutor

tofuatjava:
as I am the maintainer of thinBackup I want to publish some commits. I'm not able to execute these script myselve because of an NPE. However, it would be cool if I could puplish again...
Thanks
Thomas

### DIFF
--- a/permissions/plugin-SCTMExecutor.yml
+++ b/permissions/plugin-SCTMExecutor.yml
@@ -1,6 +1,6 @@
 ---
-name: "sctmexecutor"
+name: "SCTMExecutor"
 paths:
 - "hudson/plugins/sctmexecutor/SCTMExecutor"
 developers: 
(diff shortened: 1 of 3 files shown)


daniel-beck:
> I'm not able to execute these script myselve because of an NPE. 

@tofuatjava No worries, you're not supposed to. It requires Artifactory admin permissions.

daniel-beck:
@tofuatjava I'm not sure the syntax of these files is correct. In any case, it's inconsistent with all others – could you please use the format shown in the README, with the user name on its own line, after a dash?

tofuatjava:
@daniel-beck - changed the format as requested.

daniel-beck:
@tofuatjava FYI I added a few commits changing some minor details.

PR #121, opened 2016-11-15
zap-plugin release authorization permission request

JordanGS:
Created my authorization request to get the ball rolling, the repository was created but there are some issues since i am no longer the administrator of it. If someone could help that would be great.

https://github.com/jenkinsci/zap-plugin

If you need anything more please let me know, thanks 👍

### DIFF
--- a/permissions/plugin-zap.yml
+++ b/permissions/plugin-zap.yml
@@ -1,0 +1,6 @@
+---
+name: "zap"
+paths:
+- "org/jenkins-ci/plugins/zap"
+developers:
+- "jordangs"


daniel-beck:
Wrong wiki URL in the POM will result in distribution being suppressed. Plugins need valid wiki URLs.

permissions/plugin-zap.yml
```diff
@@ -0,0 +1,6 @@
+---
+name: "zap"
+paths:
+- "org/jenkinsci/plugins/zap"
```
daniel-beck: Why did you strip the hyphen from this? Note that only the package name cannot handle the dash.

daniel-beck:
Other than surprising missing dashes in the POM, looks good. You can of course keep the broken "reverse URL" groupId in the POM if you like it, but looks accidental rather than deliberate.

JordanGS:
@daniel-beck clarification please.

My actual path is `zap/src/main/java/org/jenkinsci/plugins/zap`

because i cannot have a hyphen as part of the path name.

The yml should however say

```
paths:
- "org/jenkins-ci/plugins/zap"
```

is that correct?

daniel-beck:
@JordanGS 

Yes.

> Note that only the package name cannot handle the dash.

Package names and groupId are unrelated, technically. They just start typically with the same prefix based on a similar convention, and packages don't allow hyphens.

JordanGS:
@daniel-beck updated, good clarification.

PR #123, opened 2016-11-16
Create plugin-vectorcast-coverage.yml

richardmkeeble:
Please add permissions for the vectorcast-coverage plugin


daniel-beck:
Repo: https://github.com/jenkinsci/vectorcast-coverage-plugin/blob/master/pom.xml

PR #128, opened 2016-11-23
Add minio-storage plugin

NitishT:


daniel-beck:
None of these user accounts exist.

If you haven't bothered to read this repository's README before opening a PR, now would be a great time to do that.

NitishT:
Actually I did bother to read it, but now well enough apparently. I confused Jenkins LDAP user name with GitHub user names. 

Updated the xml with Jenkins LDAP user name.

daniel-beck:
@NitishT The account has never logged in to Artifactory, which means this change won't be effective until some time *after* the first login – which may well mean the first upload will fail. SO I recommend logging in now.

daniel-beck:
@NitishT Also, if you have advice on how to prevent user name confusion like yours in the readme, I'd appreciate pointers.

NitishT:
I remember logging in with the account to the Artifactory before raising the PR. But just in case, I logged in with the account again.

Regarding the user name confusion, I think it is clearly mentioned in the README, it was more of an oversight on my end.

To make it clearer, the sentence 

> A set of user names (Jenkins community user accounts in LDAP, the same as used for wiki and JIRA) allowed to upload this artifact to Artifactory

can be probably reframed to

> A set of user names ([jenkins-ci.org](https://wiki.jenkins-ci.org/display/JENKINS/User+Account+on+Jenkins) account user names used to login to JIRA and wiki) allowed to upload this artifact to Artifactory

daniel-beck:
> jenkins-ci.org account user names

I'm no longer using that term as we're in the process of moving domains over to jenkins.io ;-)

PR #137, opened 2016-11-30
Requesting permissions for onesky plugin

ricardoespsanto:


ricardoespsanto:
Hosting request has already been approved: https://issues.jenkins-ci.org/browse/HOSTING-236 & plugin is here: https://github.com/jenkinsci/onesky-plugin

PR #154, opened 2016-12-09
Adding BlackDuck hub plugin permission

jrichardBD:
# Description

This is a Jenkins plugin that allows Users to run BlackDuck Hub scans in Jenkins.

https://github.com/jenkinsci/blackduck-hub-plugin

https://issues.jenkins-ci.org/browse/HOSTING-215

@daniel-beck

### DIFF
--- a/permissions/plugin-blackduck-hub.yml
+++ b/permissions/plugin-blackduck-hub.yml
@@ -1,0 +1,8 @@
+---
+name: "blackduck-hub"
+paths:
+- "com/blackducksoftware/integration/blackduck-hub"
+developers:
+- "jrichard"
+- "akamen"
+- "bdsoss"


akamen:
Thanks in advance @daniel-beck

oleg-nenashev:
👍 , but Just FYI you will unlikely be able to release the plugin with your current POM. The deployment process now supports HTTPS only. The best solution would be to update the Parent POM to 2.x latest (you can still keep the core dependency on 1.809.x)

akamen:
@oleg-nenashev  the pom has been updated to 2.x - we got past the release goals, but unable to perform the deploy due to a permission issue.

daniel-beck:
"message" : "Permission target contains a reference to a non-existing user 'akamen'."

daniel-beck:
@oleg-nenashev Please make sure to check whether users exist in Artifactory before merging. If you can't, just poke me and I'll do it. It's the most common source of problems we have with these permissions files.

jrichardBD:
The User's I included in the permission file (jrichard, bdsoss, and akamen) all exist and have all logged into the Artifactory server https://repo.jenkins-ci.org.  But we still get a forbidden error when we try to deploy the artifacts

PR #160, opened 2016-12-16
[INFRA-1005] Allow contributors to deploy -SNAPSHOTs

stephenc:
See [INFRA-1005](https://issues.jenkins-ci.org/browse/INFRA-1005)

- Also tidy up the updater and allow for dry runs and off-line testing

- Also adds "stephenconnolly" as a contributor to [git](https://github.com/jenkinsci/git-plugin) and [mercurial](https://github.com/jenkinsci/mercurial-plugin) plugins @jglick @MarkEWaite

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ node('java') {
         stage 'Run'
         withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'artifactoryAdmin',
                           usernameVariable: 'ARTIFACTORY_USERNAME', passwordVariable: 'ARTIFACTORY_PASSWORD']]) {
-            sh 'target/appassembler/bin/repository-permissions-updater -d $PWD/permissions -w $PWD/json'
+            sh 'target/appassembler/bin/repository-permissions-updater --definitionDir $PWD/permissions --workDir $PWD/json'
         }
     } finally {
         stage 'Archive'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,11 +30,7 @@ node('java') {
         stage 'Run'
         withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'artifactoryAdmin',
                           usernameVariable: 'ARTIFACTORY_USERNAME', passwordVariable: 'ARTIFACTORY_PASSWORD']]) {
-            sh '${JAVA_HOME}/bin/java' +
-                    ' -DdefinitionsDir=$PWD/permissions' +
-                    ' -DartifactoryApiTempDir=$PWD/json' +
-                    ' -Djava.util.logging.SimpleFormatter.format="%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$s: %5$s%6$s%n"' +
-                    ' -jar target/repository-permissions-updater-*-bin/repository-permissions-updater-*.jar'
+            sh 'target/appassembler/bin/repository-permissions-updater -d $PWD/permissions -w $PWD/json'
         }
     } finally {
         stage 'Archive'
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -30,11 +30,7 @@ node('java') {
         stage 'Run'
         withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'artifactoryAdmin',
                           usernameVariable: 'ARTIFACTORY_USERNAME', passwordVariable: 'ARTIFACTORY_PASSWORD']]) {
-            sh '${JAVA_HOME}/bin/java' +
-                    ' -DdefinitionsDir=$PWD/permissions' +
-                    ' -DartifactoryApiTempDir=$PWD/json' +
-                    ' -Djava.util.logging.SimpleFormatter.format="%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$s: %5$s%6$s%n"' +
-                    ' -jar target/repository-permissions-updater-*-bin/repository-permissions-updater-*.jar'
+            sh 'target/appassembler/bin/repository-permissions-updater -d $PWD/permissions -w $PWD/json'
```
daniel-beck: I strongly prefer long options for readability.

README.md
```diff
@@ -52,6 +52,10 @@ Create a new YAML file similar to existing files.
 
 Edit the `developers` list in the YAML file for the plugin.
 
+### Adding a `-SNAPSHOT` only uploader to an existing plugin
+
+Create or edit the `contributors` list in the YAML file for the plugin.
```
daniel-beck: A lot more of the readme needs to be updated to incorporate this properly.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -288,9 +323,58 @@ public class ArtifactoryPermissionsUpdater {
...
+                            System.getenv("ARTIFACTORY_PASSWORD").toCharArray());
+                }
+            });
+        }
+
+        File definitionsDir = new File(options.d?:defDefnDir)
+        File workDir = new File(options.w?:defWorkDir)
+        generateApiPayloads(definitionsDir, workDir)
+        submitPermissionTargets(workDir, dryRun)
+        if (!offline) {
```
daniel-beck: Why this condition if offline implied dry-run?

daniel-beck:
@stephenc I wonder whether snapshot permission could be granted to anyone. Since snapshots are not distributed further, and referenced downstream using their exact date, even malicious uploads wouldn't be a problem AFAICT. WDYT?

stephenc:
@daniel-beck snapshot deployment for all would be super-awesome... I'm just trying to unblock my small issue with least change

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
stephenc: dry run when on-line will tell you the permissions for removal.

the submit doesn't care about on-line once dryRun is enabled, but the clean up needs auth

MarkEWaite:
+1 for @stephenc being added as a contributor to the git plugin.  Should he also be added as a contributor to the git client plugin during this same change.

daniel-beck:
@stephenc I granted unrestricted upload permissions to `snapshots` to you, temporarily.

Longer term, I think I can open up snapshot upload permissions in general, as I doubt there's a security problem with that. Will need to think about it first though.

batmat:
I tend to agree this should not hurt, and would test some things indeed. Or alternatively we could have PR build going up to deploy.

daniel-beck:
As written on the dev list, I opened up snapshot deployment to all users.

I may still use some of the other changes in this PR when setting up real PR builds (which unfortunately would require Artifactory admin permissions right now to find existing users accounts).

PR #161, opened 2016-12-16
[INFRA-1005] #160 Post Merge clean-up

stephenc:
Downstream of https://github.com/jenkins-infra/repository-permissions-updater/pull/160

- Removes the assembly.xml and the use of the assembly plugin as #160 has switched to appassembler instead


daniel-beck:
As #161 I'll reuse some of this when tackling PR builds, but since snapshot deployment is now open to all users, not urgent.

PR #175, opened 2017-01-11
Update plugin-notification.yml

hagzag:
_(This PR template only applies to permission changes -- ignore for changes to the tool)_

# Description
@hagzag
Add cohencil as new developer of this plugin -> https://github.com/jenkinsci/notification-plugin

### DIFF
--- a/permissions/plugin-notification.yml
+++ b/permissions/plugin-notification.yml
@@ -4,3 +4,4 @@ paths:
 - "com/tikalk/hudson/plugins/notification"
 developers:
 - "hagzag"
+- "cohencil"


orrc:
Looks good, but make sure that @cohencil follows this line from the checklist, otherwise they won't get access:
> All newly added users have logged in to Artifactory at least once

daniel-beck:
> otherwise they won't get access

FTR, as of right now, this is not a problem (anymore).

cohencil:
Thanks @daniel-beck, I made sure to follow all checklist items in advance.

PR #177, opened 2017-01-13
[JENKINS-39355] Allow stephenc to release mercurial-plugin

jglick:
Allow @stephenc to release https://github.com/jenkinsci/mercurial-plugin/pull/92 and similar as nonbeta on Monday when I may not be available.

@reviewbybees


stephenc:
:+1:

PR #197, opened 2017-01-24
Add myself as a developer

escoem:
# Description

Add myself as a developer

id: escoem

cc. @oleg-nenashev 

- https://github.com/jenkinsci/workflow-remote-loader-plugin


oleg-nenashev:
Confirmed

PR #15, opened 2016-08-28
add Joseph 'casz' Petersen

casz:
as new maintainer of Accurev plugin I would like to be added :tada:

PR #211, opened 2017-01-30
Remove old permissions from Accurev plugin

casz:
As suggested in https://github.com/jenkins-infra/repository-permissions-updater/pull/15#issuecomment-242981574
Should we remove the old ones?

Also is `org/jvnet/hudson/plugins/accurev` still relevant?

https://github.com/jenkinsci/accurev-plugin

### DIFF
--- a/permissions/plugin-accurev.yml
+++ b/permissions/plugin-accurev.yml
@@ -2,6 +2,5 @@
 name: "accurev"
 paths:
 - "org/jenkins-ci/plugins/accurev"
-- "org/jvnet/hudson/plugins/accurev"
 developers:
 - "casz"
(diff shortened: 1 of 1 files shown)


daniel-beck:
>  Should we remove the old [ones]?

Makes sense to me. Pinging @rym002 just in case so they can object. It doesn't look like @ndeloof was ever really a maintainer.

> Also is `org/jvnet/hudson/plugins/accurev` still relevant?

If you don't plan on ever again publishing a release with this group/artifactId, it can be removed.

daniel-beck:
CC @orrc

casz:
> > Also is org/jvnet/hudson/plugins/accurev still relevant?
> If you don't plan on ever again publishing a release with this group/artifactId, it can be removed.
Last release to org.jvnet.hudson.plugins.accurev was 2011. So no I don't plan on that.
https://mvnrepository.com/artifact/org.jvnet.hudson.plugins/accurev

daniel-beck:
Looks good.

I'll leave this PR open for a bit so @rym002 can still speak up if interested.

casz:
Sounds good 👍

PR #218, opened 2017-02-06
@aheritier should be able to release lib-jenkins-maven-embedder

oleg-nenashev:
Maven is his baby at the end of the day


oleg-nenashev:
@aheritier is fine with it

PR #219, opened 2017-02-06
add bparees to openshift plugins

gabemontero:
@jenkinsadmin or others PTAL

In addition to adding @bparees to openshift-login and openshift-pipeline, we noticed that at 
https://github.com/jenkinsci/openshift-pipeline-plugin/settings/collaboration "Everyone" seems to have write
access.

However, when we look at say https://github.com/jenkinsci/openshift-client-plugin/settings/collaboration or https://github.com/jenkinsci/openshift-login-plugin/settings/collaboration that is not the case.

Any idea how we can remove "Everyone" from having write access to openshift-pipeline?

thanks

### DIFF
--- a/permissions/plugin-openshift-login.yml
+++ b/permissions/plugin-openshift-login.yml
@@ -4,3 +4,4 @@ paths:
 - "org/openshift/jenkins/openshift-login"
 developers:
 - "gmontero"
+- "bparees"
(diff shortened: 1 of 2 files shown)


daniel-beck:
> Any idea how we can remove "Everyone" from having write access to openshift-pipeline?

By asking. Done.

gabemontero:
Perfect.  Thanks @daniel-beck !

PR #222, opened 2017-02-10
Create component-lib-version-number.yml

stephenc:
Found out that we have no permissions file for https://github.com/jenkinsci/lib-version-number

@jglick @daniel-beck @reviewbybees

### DIFF
--- a/permissions/component-lib-version-number.yml
+++ b/permissions/component-lib-version-number.yml
@@ -1,0 +1,8 @@
+---
+name: "version-number"
+paths:
+- "org/jenkins-ci/version-number"
+developers:
+- "jglick"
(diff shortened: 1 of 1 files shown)


permissions/component-lib-version-number.yml
```diff
@@ -0,0 +1,8 @@
+---
+name: "version-number"
```
jglick: Is this supposed to match the `artifactId` I guess?

daniel-beck: Yes.

daniel-beck:
File should be named `component-version-number.yml`, otherwise looks good.

PR #285, opened 2017-04-19
add rymurphy/bparees as needed to openshift plugins

gabemontero:
updating developer list for openshift plugins

@jenkinsadmin PTAL

### DIFF
--- a/permissions/plugin-openshift-client.yml
+++ b/permissions/plugin-openshift-client.yml
@@ -5,3 +5,5 @@ paths:
 developers:
 - "jupierce"
 - "gmontero"
+- "bparees"
+- "rymurphy"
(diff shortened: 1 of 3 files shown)


gabemontero:
@oatmealraisin @bparees fyi

bparees:
are we both on the login plugin?

gabemontero:
you are ... @oatmealraisin isn't ... thought about it and then went conservative, but can add him if you like.

bparees:
it's fine as is.

daniel-beck:
`rymurphy` hasn't logged into Artifactory, so the entire change will only be applied when they do.

oatmealraisin:
@daniel-beck sorry, I have logged in now. Thanks for the notice!

daniel-beck:
Permission changes got applied.

PR #324, opened 2017-05-30
Provide a way to check the Jenkins ID does exist

batmat:
Ping @daniel-beck @oleg-nenashev


daniel-beck:
Except for the few users who don't have Jira account because something went wrong in account sync…

@rtyler recently set up the reports host ([INFRA-947](https://issues.jenkins-ci.org/browse/INFRA-947), so we can now generate a list of known Artifactory user names on trusted.ci and dump them there. I'd prefer that approach unless there's opposition to enumerating user names – but since anyone can get the user names who ever uploaded something, I wouldn't consider that a big deal TBH.

batmat:
OK, fair enough. This would only improve the situation for a portion of people. Not necessary if INFRA-947 is more than a thought indeed.

PR #337, opened 2017-06-15
Add @stephenc as a maintainer of the structs plugin

stephenc:
@jglick or @oleg-nenashev

### DIFF
--- a/permissions/plugin-structs.yml
+++ b/permissions/plugin-structs.yml
@@ -6,3 +6,4 @@ developers:
 - "jglick"
 - "kohsuke"
 - "oleg_nenashev"
+- "stephenconnolly"


stephenc:
@daniel-beck PING

daniel-beck:
@stephenc I periodically (multiple times per day) review incoming notifications, that include all repos I watch. 3 minutes is not enough times to require a followup ping.

stephenc:
@daniel-beck it's more I have a release half cut: https://github.com/jenkinsci/structs-plugin/tree/structs-parent-1.8 and I would like to get it finished

oleg-nenashev:
Not sure if we need to wait for @jglick , who is the real maintainer of this plugin

stephenc:
So the reason I started the release was that @jglick told me I could release 1.8 myself... so I -
 like the fool that I am - assumed he had checked and I was one of the maintainers... turns out I wasn't but I'm assuming he's ok

daniel-beck:
Given the restrictive wording of https://github.com/jenkinsci/structs-plugin/pull/21#issuecomment-308267072 I'd rather not assume his approval.

@stephenc Do you have a better reference than that?

stephenc:
@daniel-beck well he said it in a call he and I had, so if you want to wait for @jglick that's fine... hopefully my laptop does not die between now and then

daniel-beck:
Good enough for me.

PR #341, opened 2017-06-19
Add @oleg-nenashev to the developer list of Filesystem SCM plugin

oleg-nenashev:
Request: https://groups.google.com/forum/#!topic/jenkinsci-dev/0wVGYwaoGZA

CC @kutzi and @samngms just in case they see the notification. I am ready to wait for another week just in case


daniel-beck:
Given the inactivity of the plugin and former maintainers, this request is granted per our usual policy on abandoned plugins.

PR #344, opened 2017-06-22
Enable multiple names to be specified in single permission file

olivergondza:
I was working on something more complex but it let nowhere at least pushing what needed to change to get me started.


src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -15,12 +15,12 @@ public class ArtifactoryPermissionsUpdater {
     /**
      * Directory containing the permissions definition files in YAML format
      */
-    private static final File DEFINITIONS_DIR = new File(System.getProperty('definitionsDir'))
+    private static final File DEFINITIONS_DIR = new File(System.getProperty('definitionsDir', './permissions'))
```
olivergondza: Groovy fails to identify matching `File` constructor when the value is `null`.

daniel-beck:
> something more complex

@olivergondza The PR title gives it away :-)

PR #349, opened 2017-06-27
component-jenkins-test-harness + olamy

olamy:


daniel-beck:
I don't think we have a maintainer of this per se, @jglick comes closest perhaps?

olamy:
@daniel-beck @jglick I don't mind being or not in the file. I just will be happy about a release with this change https://github.com/jenkinsci/jenkins-test-harness/commit/67e38f5d8ee18ce673c3bd0ce39e2423ceb836e1
Thanks!

daniel-beck:
@jglick PTAL -- looks like a release would be useful?

@olamy Please ping me in a day if Jesse hasn't responded.

oleg-nenashev:
I have thought it has been already released :(
Unfortunately I am also not in the list hence I cannot spin it as well.
CC @olivergondza @andresrc @stephenc could somebody release it?

PR #351, opened 2017-06-28
Add @oleg-nenashev to the List of SVN plugin developers

oleg-nenashev:
I am doing some reviews there to decrease the technical debt, and I can do releases sometimes till somebody adopts the plugin.

WDYT @stephenc @jglick ?


daniel-beck:
Origin branches make me a sad panda 😭

recena:
@stephenc I'd like to participate on Subversion Plugin but lately I don't have time.

@oleg-nenashev 👍

PR #374, opened 2017-07-20
winstone += olamy

olamy:
Did some work to introduce HTTP/2 feature so will be happy to release it :-)


oleg-nenashev:
I am in the maintainer list, so my approval is enough.
It's a shared-ownership core component

PR #385, opened 2017-07-28
Permissions for jira-finder plugin

Holmistr:
https://issues.jenkins-ci.org/browse/HOSTING-336


daniel-beck:
https://github.com/jenkinsci/jira-finder-plugin

daniel-beck:
https://github.com/jenkinsci/jira-finder-plugin/blob/c3878e9e79ac817a835b5810a908786983c529bb/pom.xml#L32 should probably be changed as well.

PR #397, opened 2017-08-10
Edit plugin path

lbordowitz:
The groupid has been updated/modernized for this plugin to allow pipeline stuff. The path permissions should reflect this.

https://wiki.jenkins.io/display/JENKINS/LabeledTestGroupsPublisher+Plugin

### DIFF
--- a/permissions/plugin-labeled-test-groups-publisher.yml
+++ b/permissions/plugin-labeled-test-groups-publisher.yml
@@ -1,6 +1,6 @@
 ---
 name: "labeled-test-groups-publisher"
 paths:
-- "org/jvnet/hudson/plugins/labeled-test-groups-publisher"
+- "org/jenkins-ci/plugins/labeled-test-groups-publisher"
 developers:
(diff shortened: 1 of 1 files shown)


daniel-beck:
Reference: https://github.com/jenkinsci/labeled-test-groups-publisher-plugin/commit/2654871928575f84d2166eb84d810504bf449910#diff-600376dffeb79835ede4a0b285078036R30

daniel-beck:
@lbordowitz Note that the vast majority of your users are way past that 1.577 already:
http://stats.jenkins.io/pluginversions/labeled-test-groups-publisher.html

Additionally, any release older than 2.46.2 has major security issues:
https://jenkins.io/security/advisory/2017-04-26/

In general, we recommend a [recent 2.x parent POM](https://github.com/jenkinsci/plugin-pom/blob/master/CHANGELOG.md) (which are versioned independently of core releases), and declare a `jenkins.version` at least in the 1.600s, for example 1.642.1.

PR #431, opened 2017-09-11
see HOSTING-409

ndeloof:
https://issues.jenkins-ci.org/browse/HOSTING-409


daniel-beck:
The PR checklist exists for a reason. Please use it as intended (i.e. keep it in the description, and check off the relevant items) in the future.

ndeloof:
@daniel-beck the only relevant element of the checklist was reference to the Jira issue, which is here. What else is missing ?

daniel-beck:
@ndeloof The checklist itself is missing, and the link to the forked repo. Both are useful to keep down the time it requires to review these requests. Note that almost all administrative duties for the Jenkins project are borne by very few contributors during their spare time, so the effort quickly adds up.

daniel-beck:
Repo: https://github.com/jenkinsci/docker-java-api-plugin

PR #438, opened 2017-09-19
add mail lists link for issue discuss

jiafu1115:
add mail lists link for issue discuss


jiafu1115:
no issue tab and no method to ask question. So just add link for more persons.

daniel-beck:
Not sure this is useful. Anyone getting here got here through https://wiki.jenkins.io/display/JENKINS/Hosting+Plugins which already has troubleshooting instructions. We'd prefer people investigate themselves first due to our limited capacity to address issues individually, so mailing lists should not be the first step.

PR #467, opened 2017-10-11
Grant Swarm plugin release permissions to @oleg-nenashev

oleg-nenashev:
Just to follow-up on the discussion with @mindjiver. I would like to release the current changes in the plugin and then to upgrade Remoting a bit more. This permissions will enable me to release the plugin on my own if required.

CC @daniel-beck


mindjiver:
👍  welcome onboard

PR #473, opened 2017-10-16
Grant @wolfs permissions for depgraph-view-plugin

wolfs:
# Description

I would like to be able to release https://github.com/jenkinsci/depgraph-view-plugin again.
I am the maintainer and the original author of the plugin.


daniel-beck:
Right, I only went back ~3 years when I initialized permissions.

PR #525, opened 2017-12-12
Grant @oleg-nenashev rights to release the label-verifier plugin

oleg-nenashev:
I am the maintainer, the plugin just didn't need much love last years


daniel-beck:
Looks like it from https://github.com/jenkinsci/label-verifier-plugin/commits/master

PR #605, opened 2018-02-22
Grant @oleg-nenashev permissions to release the JDepend plugin

oleg-nenashev:
It is a follow-up to the discussion with @ndeloof in https://github.com/jenkinsci/jdepend-plugin/pull/2 . He approved the release in that thread.


oleg-nenashev:
After the release I will mark the plugin for adoption

PR #607, opened 2018-02-26
Add LoadBalance Plugin

LinuxSuRen:
Hosting Ticket : https://issues.jenkins-ci.org/browse/HOSTING-510
Repo : https://github.com/jenkinsci/agent-loadbalance-plugin

### DIFF
--- a/permissions/plugin-agent-loadbalance.yml
+++ b/permissions/plugin-agent-loadbalance.yml
@@ -1,5 +1,5 @@
 ---
-name: "loadbalance"
+name: "agent-loadbalance"
 paths:
 - "com/surenpi/jenkins/agent-loadbalance"
 developers:
--- a/permissions/plugin-agent-loadbalance.yml
+++ b/permissions/plugin-agent-loadbalance.yml
@@ -1,6 +1,6 @@
 ---
 name: "loadbalance"
 paths:
-- "com/surenpi/jenkins/loadbalance"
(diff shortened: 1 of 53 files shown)


permissions/plugin-agent-loadbalance.yml
```diff
@@ -0,0 +1,6 @@
+---
+name: "loadbalance"
```
daniel-beck: Wrong name.

oleg-nenashev:
As a maintainer of Job Restrictions Plugin, I am against the current implementation: https://github.com/jenkinsci/agent-loadbalance-plugin/blob/master/src/main/java/com/surenpi/jenkins/loadbalance/LoadBalanceRestriction.java . It is a hack, LoadBalancer should be used instead as we discussed in the hosting request. 

I do not block hosting of the plugin in the current state, but I will reject to triage any Job Restriction issues if this plugin is installed.

permissions/plugin-agent-loadbalance.yml
LinuxSuRen: Already fixed. Thanks.

LinuxSuRen:
@oleg-nenashev Totally understand. But in that situation, it must use restrction not balancer. Maybe I should make a change, 'ResourceRestriction' could be better than 'LoadBalanceRestriction'.

PR #635, opened 2018-03-14
Throw exceptions on validation errors

daniel-beck:
This will fail PR builds with problems like wrongly named or malformed YAML files, instead of hiding a warning in the build log.


daniel-beck:
Build failure immediately discovering problem with existing file 👍

PR #638, opened 2018-03-16
[HOSTING-525] Add permissions for kubernetes-credentials-provider

jtnord:
For some reason the permissions were not set on artifactory as a result
of the hosting process request, so filing manually...

# Description

https://issues.jenkins-ci.org/browse/HOSTING-525
https://github.com/jenkinsci/kubernetes-credentials-provider-plugin

### Always

- [ ] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

### DIFF
--- a/permissions/plugin-kubernetes-credentials-provider.yml
+++ b/permissions/plugin-kubernetes-credentials-provider.yml
@@ -1,0 +1,6 @@
+---
+name: "kubernetes-credentials-provider"
+paths:
+- "com/cloudbees/jenkins/plugins/k8sCredentialProvider"
+developers:
+- "teilo"


permissions/plugin-kubernetes-credentials-provider.yml
```diff
@@ -0,0 +1,6 @@
+---
+name: "kubernetes-credentials-provider"
+paths:
+- "com/cloudbees/jenkins/plugins/k8sCredentialProvider"
```
batmat: urgh, using not only lowercase is looking for pain IMO... I guess it's too late anyway

batmat:
> For some reason the permissions were not set on artifactory as a result of the hosting process request, so filing manually...

There's no automation for this yet, so this is _normal_.

permissions/plugin-kubernetes-credentials-provider.yml
jtnord: not tool late there is no release.  
And well even after I can change it as no data is persisted, nor does this constitute a public API.

that said `k8s_credential_provider` would be the convention that is not really any better.

daniel-beck:
This will not work as explained in https://github.com/jenkins-infra/repository-permissions-updater#managing-permissions

daniel-beck:
I don't understand the redundancy in the groupId. Why not just `com.cloudbees.plugins`? https://github.com/jenkinsci/kubernetes-credentials-provider-plugin/blob/master/pom.xml#L35...L36

Also, regarding

> not tool late there is no release.
> And well even after I can change it as no data is persisted, nor does this constitute a public API.

Group IDs are ignored by Jenkins.

jtnord:
> This will not work as explained 

What won't work, I don't see what you are pointing out?

> I don't understand the redundancy in the groupId. Why not just com.cloudbees.plugins? 

because you are assuming that the only thing CloudBees has plugins for is Jenkins, and the only thing we do for Jenkins is plugins...  and we do not want collisions across different maven repos.

>Also, regarding
>> not tool late there is no release.
>> And well even after I can change it as no data is persisted, nor does this constitute a public API.
>
> Group IDs are ignored by Jenkins.


Yeah but `groupId`  *should* match the plugins base `package` in the source code, and I was referring to that.

daniel-beck:
> because you are assuming that the only thing CloudBees has plugins for is Jenkins, and the only thing we do for Jenkins is plugins... and we do not want collisions across different maven repos.

Right, missed that part -- but still `com.cloudbees.jenkins.plugins` makes sense.

> What won't work, I don't see what you are pointing out?

"Paths" is not `groupId`. You are currently allowed to upload files matching `com/cloudbees/jenkins/plugins/k8sCredentialProvider/*/kubernetes-credentials-provider-*`. I doubt this is what you want.

permissions/plugin-kubernetes-credentials-provider.yml
batmat: do we really even need such a specific groupId? Why not just `com.cloudbees.jenkins.plugins` like Daniel is saying too below?

PR #672, opened 2018-04-19
[INFRA-1571] Including information on associated GitHub repositories

jglick:
A component of [INFRA-1571](https://issues.jenkins-ci.org/browse/INFRA-1571), under discussion with @daniel-beck & @rtyler.

I plan to write some sort of script to add the `github` field to all existing components based on scraping `pom.xml` information, but this could be merged before that without harm I think—it just means that Incrementals deployment is only possible for repositories which have already been updated here.


src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -115,9 +116,11 @@ public class ArtifactoryPermissionsUpdater {
         }
 
         if (apiOutputDir.exists()) {
-            throw new IOException(apiOutputDir.path + " already exists")
+            println "$apiOutputDir.path already exists" // not fatal
```
daniel-beck: This intends to ensure the correct removal of permissions for components whose files were removed since the previous execution. Otherwise those might be retained indefinitely if the environment (workspace deletion) is not properly set up.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -180,6 +200,9 @@ public class ArtifactoryPermissionsUpdater {
      */
     private static void submitPermissionTargets(File jsonApiFileDir) {
         jsonApiFileDir.eachFile { file ->
+            if (file.name == 'github-index.json') {
```
daniel-beck: Would prefer a more magical file name that cannot conceivably clash with a plugin name. `github.index.json` or `_github-index.json` would do it.

README.md
```diff
@@ -33,16 +33,26 @@ These prefixes, like the rest of the file name, have no semantic meaning and jus
 Each file contains the following in [YAML format](https://en.wikipedia.org/wiki/YAML):
 
 - A `name` (typically mirrored in the file name), this is also the `artifactId` of the Maven artifact.
+- A `github` field indicating the GitHub organization and repository which is expected to produce these artifacts.
```
daniel-beck: Add *(optional)*

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
daniel-beck: In fact, it does not seem this file should even be here. AFAIR this directory contains the API payloads for Artifactory, so this should be stored elsewhere.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
jglick: OK, can revert. This just made it a lot more convenient to iteratively test.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
jglick: > a more magical file name that cannot conceivably clash with a plugin name

I can try to come up with something, on the off chance someone decides to write a plugin named `github-index`.

> this should be stored elsewhere

That then requires a new input parameter for a new temporary directory. And a patch to `Jenkinsfile`, which as a submitter without write permission to the repository, I cannot then test. Seems like a hassle.

README.md
jglick: _Currently_ it is optional. I would like to make it mandatory, as soon as I have time to write up a script to scan all POMs and the like in a @jenkinsci checkout. Marking it optional would indicate to people submitting new files that they could skip this step, when I would want them to include the field.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
daniel-beck: > a new input parameter for a new temporary directory

Probably easy enough to do downstream later? Or is your use of this tied to specific file paths?

README.md
daniel-beck: You could re-use the `scm` attributes in http://updates.jenkins-ci.org/update-center.actual.json (which is also where the plugin site gets its GitHub links).

Not always present because POMs can be surprisingly terrible, but already does some guesswork in https://github.com/jenkins-infra/update-center2/blob/91c94226b2f9b3428919ac929b33b76b2139a8d1/src/main/java/org/jvnet/hudson/update_center/Plugin.java#L335 and related code.

jglick: Thanks, that is a good tip.

jglick:
W.r.t. the name of the index file:

> is your use of this tied to specific file paths?

https://github.com/jenkins-infra/community-functions/pull/8 expects a specific artifact path. Obviously that can be changed if a new one is picked. But again I cannot personally _test_ any change which would require modifications to `Jenkinsfile`.

PR #680, opened 2018-04-27
Some github fields useful for JEP-305 application to JEP-202

jglick:
Some metadata continuing #672, pending an exhaustive list.


jglick:
Trying to fix a build failure in https://github.com/jenkinsci/workflow-api-plugin/pull/67 for example.

PR #687, opened 2018-05-08
[INFRA-1598] Adding github fields

jglick:
[INFRA-1598](https://issues.jenkins-ci.org/browse/INFRA-1598): follows up #672.

- [X] Add SCM information from UC as suggested by @daniel-beck
- [x] Add more information from POM analysis


src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -141,10 +141,6 @@ public class ArtifactoryPermissionsUpdater {
                     pathsByGithub[definition.github] = paths
                 }
                 paths.addAll(definition.paths)
-            } else {
-                /* TODO once all have been filled in
-                throw new IOException("$definition.name does not specify the `github` field")
-                */
```
batmat: Wondering: should we better introduce an exception list to handle this, and this way be able to check this for new and modified in the future?

daniel-beck:
👎 Strongly opposed to supporting plugins hosted outside the jenkinsci organization this way. As far as I'm concerned, they have no source code hosting. We haven't allowed new plugins to do this in several years, it was never project policy to allow it -- it just was not actively enforced due to lack of permissions.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
jglick: We could, though it is probably overkill. The instructions include the field, so I would hope newly filed entries include it. For the moment the only impact of forgetting it is that you cannot deploy to Essentials until you add it.

jglick:
@daniel-beck in #jenkins-infra:

> the problem with some repos will be that actual development happens elsewhere and the jenkinsci repos are abandoned
> so if your expectation is to build on top of this metadata, I'd be surprised if that worked without notable problems

Currently this metadata is used by `incrementals-publisher`: _if_ you have a `Jenkinsfile` in your repo causing it to be built in the [standard place](https://ci.jenkins.io/job/Plugins/), _and_ your repo has been incrementalified, then your PRs are eligible for automatic deployment. If a repo is development elsewhere, then these preconditions will not hold, so it does not really matter. But I preferred to update metadata in bulk and set the expectation that new additions going forward will include it, so that it does not need to be added on a case-by-case basis.

batmat:
@daniel-beck is this acceptable to merge now for you? I depend on this to deploy https://github.com/jenkinsci/essentials-plugin/ in incrementals.

I can file a dedicated PR, but this will conflict this one. So if the fix @jglick did addresses your concern, I would rather merge here and do it in one go.

Thanks!

batmat:
Same for https://issues.jenkins-ci.org/browse/JENKINS-51284 BTW

daniel-beck:
https://github.com/jglick/repository-permissions-updater/blob/github/permissions/pom-github-integration-parent.yml looks broken, even besides the fact these aren't actual repos (and again, no fork in jenkinsci – it was even migrated out).

jglick:
> I can file a dedicated PR, but this will conflict this one.

Well, if you make the same change, it will merge cleanly, FWIW.

> `pom-github-integration-parent.yml` looks broken

Though I had fixed all these but must have missed this one, will fix.

> For some cases the mapping is not fully correct.

I am sure there are some such cases, because this is the result of scripts working against sometimes incorrect or contradictory metadata coming from plugin maintainers. We can always fix particular cases in follow-up PRs.

> Some files are also not updated, e.g. Yet Another Docker Plugin

Ditto, some plugins just have weird or inconsistent metadata. This PR attempts to get 95% of the stuff right—if you did nothing wrong or unusual, your plugin will be listed.

daniel-beck:
These source code locations look wrong (e.g. copy & paste):

	permissions/plugin-dotnet-as-script.yml:
	    3: github: "jenkinsci/plugin-pom"

	permissions/plugin-rocketchatnotifier.yml:
	    3: github: "jenkinsci/plugin-pom"

	permissions/plugin-drecycler.yml:
	    3: github: "jenkinsci/diskcheck-plugin"

	permissions/plugin-maven-deployment-linker.yml:
	    3: github: "jenkinsci/favorite-plugin"

	permissions/component-jenkins-maven-plugin.yml:
	    3: github: "jenkinsci/maven-repository-plugin"

	permissions/plugin-repository.yml:
	    3: github: "jenkinsci/maven-repository-plugin"

	permissions/plugin-python.yml:
	    3: github: "jenkinsci/scp-plugin"

	permissions/component-library.yml:
	    3: github: "jenkinsci/findbugs-plugin"

	permissions/plugin-embeddable-build-status.yml:
	    3: github: "jenkinsci/git-plugin"

	permissions/plugin-pry.yml:
	    3: github: "jenkinsci/git-plugin"

Non-canonical (301) or non-existing (404) source locations:

	jenkinsci/archived-artifact-url-viewer
	jenkinsci/assembla-merge-request-builder
	jenkinsci/autocomplete-parameter
	jenkinsci/cloudfoundry
	jenkinsci/cluster-stats
	jenkinsci/codebeamer-coverage-publisher
	jenkinsci/collabnet-uc4-deploy-plugin
	jenkinsci/copy-data-to-workspace-plugin-plugin
	jenkinsci/customtools-plugin
	jenkinsci/gerrit-verify-status-reporter
	jenkinsci/google-source-plugin-plugin
	jenkinsci/groovy-events-listener-plugin-master
	jenkinsci/jamm-plugin
	jenkinsci/JDK_Parameter_Choice
	jenkinsci/jenkins-ignore-committer-strategy-plugin
	jenkinsci/jenkins-keycloak-plugin
	jenkinsci/jenkins-xcp-ci-plugin
	jenkinsci/keep-slave-offline
	jenkinsci/m2release-plugin/distfork-plugin
	jenkinsci/memegen
	jenkinsci/office-365-connector
	jenkinsci/oki-docki-plugin
	jenkinsci/seleniumhtmlreport
	jenkinsci/selfie-trigger-plugin-plugin
	jenkinsci/slave-utilization-plugin
	jenkinsci/ssh2easy
	jenkinsci/suite-test-groups-publisher-plugin
	jenkinsci/validating-string-parameter-pluginn
	jenkinsci/vsts-cd-plugin-plugin
	jenkinsci/XL+Release-plugin
	jenkinsci/xml-job-to-dsl-plugin

While redirects to other repos after a rename should be acceptable in the future, for this initial definition these should be the canonical locations. Especially since some repos may have been moved outside the jenkinsci organization.

daniel-beck:
Ideally we'd just use a new org like `@jenkins-plugins`, migrate all repos there, and name them identical to the artifactId, with/without suffix, whatever. Would make this entire thing unnecessary…

daniel-beck:
@jglick You forgot `plugin-xlrelease-plugin.yml`

daniel-beck:
Still concerned about the correctness of this given the numerous issues we've discovered, but if we act accordingly and fix mistakes as needed, sure.

PR #689, opened 2018-05-08
Update plugin-configuration-as-code.yml

ndeloof:
updated groupId for configuration-as-code

# Description

https://github.com/jenkinsci/configuration-as-code-plugin updated groupID see 
https://github.com/jenkinsci/configuration-as-code-plugin/commit/d7fe9956a5a6e571aca1f7bdee80401321de9048

### Always

### DIFF
--- a/permissions/plugin-configuration-as-code.yml
+++ b/permissions/plugin-configuration-as-code.yml
@@ -2,5 +2,7 @@
 name: "configuration-as-code"
 paths:
 - "org/jenkins-ci/plugins/configuration-as-code"
+- "io/jenkins/configuration-as-code"
 developers:
 - "praqma"
+- "ndeloof"


oleg-nenashev:
@ndeloof is listed as a developer in https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/pom.xml . I doubt it needs extra authorization @daniel-beck . But CC @ewelinawilkosz just in case

ewelinawilkosz:
If I’m the one who can authorize then consider it done! If I misunderstood then sorry :)

ndeloof:
by the way, @ewelinawilkosz I wonder the upload permission is set for user "praqma", and not individual contributors at Praqma.

daniel-beck:
Just wasn't sure how the release model in CasC was supposed to work, with @ewelinawilkosz's approval this is fine.

PR #723, opened 2018-06-07
Added praqmajutils library to jenkins artifactory

MadsNielsen:
# Description

Utility library used in ClearCase UCM library and various other plugins praqma has developed.
https://github.com/Praqma/praqmajutils

### Always

- [x] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

### DIFF
--- a/permissions/component-praqmajutils.yml
+++ b/permissions/component-praqmajutils.yml
@@ -1,0 +1,8 @@
+---
+name: "praqmajutils"
+github: "praqma/praqmajutils"
+paths:
+- "net/praqma/praqmajutils"
+developers:
+- "praqma"
+- "madsnielsen"


MadsNielsen:
Again another library we want to move. Added myself @MadsNielsen and @ReleasePraqma as publishers.

jglick:
Have you considered publishing to Central (via OSSRH) instead? This is currently using `hpi` packaging yet it does not seem to call any Jenkins APIs. If it is indeed a pure library, it would be better published independently (with `jar` packaging); the Jenkins plugin can simply declare a Maven dependency on it in order to bundle it.

MadsNielsen:
@jglick That is exact what we currently do, i changed the packaging to .hpi from .jar to make it clear it is intended for Jenkins. The plugin is already published but we're shutting down our own maven repository soon, and i thought this was the easiest way to go about it for finding a new home for this dependency. 

I'll close this if you do not allow any libraries the do not use any Jenkins libraries and i'll see if i can figure out to get it published via OSSRH.

daniel-beck:
We actually don't have an established process for these. There are some exceptions (those with `component-` file name), but AFAICT those mostly are upstream libraries we patched, or other obvious Jenkins specific libraries.

If you still want to go this route, I recommend an email to the dev list.

MadsNielsen:
@daniel-beck Yeah, that's the reason i added the `component` name in this request. I'll try to go the OSSRH route instead since i might have misunderstood what component was used for in this context in the Jenkins organisation. This particular library is used in 4 different Jenkins plugins and nowhere else. TBH i only did it here since i thought it was the fastest route to get an artifact hosted.

MadsNielsen:
@daniel-beck Alright, i think it should be clarified in the documentation that, these components are usually structural in nature, like shaded libraries, and for jenkins core, or alternatively has to use a Jenkins API in order to be eligable for hosting consideration.

oleg-nenashev:
> This particular library is used in 4 different Jenkins plugins and nowhere else.

In such case we could consider converting it to an API plugin with a generic name. It's what we did for UCM API a while ago.

But for the linked codebase I am not sure it's feasible. Praqma utils is a pack of various small classes which duplicate existing APIs in newer Java versions and logging systems. I am not sure what would be the benefit for the community

MadsNielsen:
@oleg-nenashev Yeah, it was also one of the conclusions i came to (this is when JDK7 was in beta). So i'm OK with this one being rejected for hosting, but i still maintain that the requirements are rather vauge for this kind of API. 

Im looking forward to the UCM API being accepted so we can finally move to the cloud and away from our old maven repo :)

oleg-nenashev:
@MadsNielsen I agree with your feedback. IMHO it would be useful to formalize requirements, and it would be a subject for JEP IMHO. I created https://issues.jenkins-ci.org/browse/INFRA-1662 for it

PR #269, opened 2017-03-31
Takeover crowd2 plugin

hypery2k:
# Description

I want to continue maintaining the [crowd2 plugin](https://github.com/jenkinsci/crowd2-plugin) (@integer,@theit,@KostyaSha)  We're using the plugin at our company and I would like to merge the PRs and release a new version to fix same minor bugs

# Permissions pull request checklist

### Always

- [x] Add link to plugin/component Git repository in description above

### When adding new uploaders (this includes newly created permissions files)

- [x] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)
- [x] Use the Jenkins community (LDAP) account name in the YAML file, not the GitHub account name
- [x] [All newly added users have logged in to Artifactory at least once](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

PR #724, opened 2018-06-07
Access to plugin-crowd2

pingunaut:
# Description

I want to continue maintaining the crowd2 plugin: https://github.com/jenkinsci/crowd2-plugin
Former maintainers: @integer,@theit,@KostyaSha
Approval was provided in the mailing list: https://groups.google.com/forum/#!topic/jenkinsci-dev/sFejhRvZiIM

### DIFF
--- a/permissions/plugin-crowd2.yml
+++ b/permissions/plugin-crowd2.yml
@@ -6,3 +6,4 @@ paths:
 developers:
 - "integer"
 - "mreinhardt"
+- "pingunaut"


daniel-beck:
@oleg-nenashev @pingunaut 

Premature merge given https://github.com/jenkins-infra/repository-permissions-updater/pull/269 as it doesn't look like Kostya considered himself to be a maintainer even back then. Needs to be implemented as if @hypery2k were the only maintainer on record.

daniel-beck:
@oleg-nenashev Good point, I missed that 👍  In that case this is fine.

PR #743, opened 2018-06-22
update sync plugin maintainers

gabemontero:
@daniel-beck ptal

### DIFF
--- a/permissions/plugin-openshift-sync.yml
+++ b/permissions/plugin-openshift-sync.yml
@@ -4,7 +4,5 @@ github: "jenkinsci/openshift-sync-plugin"
 paths:
 - "io/fabric8/jenkins/plugins/openshift-sync"
 developers:
-- "jimmi"
 - "gmontero"
 - "bparees"
-- "rymurphy"


oleg-nenashev:
Would be nice to have a link with justification or authorization from the removed contributors

batmat:
@gabemontero out of curiosity, why did you blank out the default template for this repository? It has actually been added for a reason: so that we all save time showing important things *were* checked...

gabemontero:
@oleg-nenashev sure thing .... the was mentioned briefly on #jenkins but yeah makes sense to list it here too.

Both Jimmi Dyson and Ryan Murphy left the OpenShift project (as well as RedHat) and are no longer contributing to this project.  As part of that, I worked with @daniel-beck on #jenkins to make me the primary contact for any Jira's opened against openshift-sync-plugin so I get the notifications, etc.  During that exchange, @daniel-beck suggested I clean up this list.

that said ... @jimmidyson @oatmealraisin - please chime if you'd prefer to be listed in this permission file, as you still might plan to contribute here.

@batmat I simply rushed through things and was not paying attention...would you all prefer I close this PR out and reopen one?

@bparees FYI

oleg-nenashev:
The template can be restored in this PR if needed. I am fine with keeping the PR as is, because the template does not help much with "remove users" use-case anyway.

Regarding the cleanup, I will defer it to @daniel-beck. AFAIK we do not have clear guidelines about removing maintainers by other maintainers.

daniel-beck:
We have no established process for this, but have accepted requests like this in the past based on the assumption that it's easy to restore access as needed in case of maintainership disputes. Pinging the removed users ensures visibility, so thanks for that.

This change looks reasonable, so I'm going to merge it.

PR #798, opened 2018-08-01
Add lvotypkova as maintainer.

lvotypko:
Discuss here https://github.com/jenkinsci/nested-view-plugin/pull/19


daniel-beck:
Ideally this takeover of maintainership would be approved by @ctapobep who apparently released the latest version, but as that's been 5 years ago, I think we can approve this after a timeout.

PR #803, opened 2018-08-10
Create plugin-configuration-as-code-support.yml

ndeloof:
# Description

https://github.com/jenkinsci/configuration-as-code-plugin/tree/master/support

### Always

cc @oleg-nenashev

### DIFF
--- a/permissions/plugin-configuration-as-code-support.yml
+++ b/permissions/plugin-configuration-as-code-support.yml
@@ -1,0 +1,8 @@
+---
+name: "configuration-as-code-support"
+github: "jenkinsci/configuration-as-code-plugin"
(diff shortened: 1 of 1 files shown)


daniel-beck:
Am I the only one confused that `configuration-as-code` is used as both an artifactId, and part of groupIds? Is this a common pattern?

ndeloof:
@daniel-beck because `configuration-as-code` initially was a standalone artifact.
nothing in maven model forbid to mix artifact and groups within the same path

oleg-nenashev:
FTR https://github.com/jenkinsci/configuration-as-code-plugin/issues/356

PR #821, opened 2018-08-22
Create plugin-synopsys-coverity.yml

jamesrichard91:
# Description
https://github.com/jenkinsci/synopsys-coverity-plugin
https://issues.jenkins-ci.org/browse/HOSTING-611

### Always

- [X] Add link to plugin/component Git repository in description above

- [X] Add link to resolved HOSTING issue in description above

- [X] Make sure the file is created in `permissions/` directory
- [X] `artifactId` (pom.xml) is used for `name` (permissions YAML file).
- [X] [`groupId` / `artifactId` (pom.xml) are correctly represented in `path` (permissions YAML file)](https://github.com/jenkins-infra/repository-permissions-updater/#managing-permissions)
- [X] Check that the file is named `plugin-${artifactId}.yml` for plugins

- [X] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)
- [X] Use the Jenkins community (LDAP) account name in the YAML file, not the GitHub account name
- [X] [All newly added users have logged in to Artifactory at least once](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

### DIFF
--- a/permissions/plugin-synopsys-coverity.yml
+++ b/permissions/plugin-synopsys-coverity.yml
@@ -1,0 +1,10 @@
+---
+name: "synopsys-coverity"
+github: "jenkinsci/synopsys-coverity-plugin"
+paths:
+- "org/jenkins-ci/plugins/synopsys-coverity"
+developers:
+- "gkillough"
+- "jrichard"
+- "akamen"
+- "bdsoss"


oleg-nenashev:
I maybe a bit too late to the party, but what is the difference from https://github.com/jenkinsci/coverity-plugin ? Using Gradle stack is something I would also advice against for the Jenkins plugins.

CC @slide

akamen:
@oleg-nenashev 
There is a substantial improvement in usability.   We would like to encourage our customers to use Jenkins, but the existing plugin does not achieve that.  Functionality is similar, but it is a complete rewrite and a lot less code :)  
As for the gradle stack, we have been using the gradle stack for our two other Jenkins plugins (blackduck and blackduck-detect), because all of our libraries are gradle based and we gain a lot of efficiency.  We were able to achieve that transition from maven, largely due to Jenkins documentation allowing for the gradle stack.

akamen:
Oleg, delighted to hear you support the refresh.  Small world!  Unfortunately a large company as Synopsys makes replacing it wholesale a difficult proposition.  Migrating the users would be very difficult, if you recall, the configuration on the old plugin is a bit....intense.  We do not want to touch that with a 10 foot pole.

So calling ours 2.0.0 is disingenuous, they are completely different plugins.   We are ok with just removing the old plugin from the marketplace to avoid confusion, as long as this new plugin is readily available and public.    If you could allow for a temporary co-existence, we will make sure the old one is removed.  Can that be done?

oleg-nenashev:
Thanks for the clarification!

>  If you could allow for a temporary co-existence, we will make sure the old one is removed. Can that be done?

Yes, it can. You are the maintainer of the original plugin, and you know what you are doing. So I am ok with the approach. IIUC @slide is also fine with that. 

My only concern that "coverity" and "synopsys-coverity" artifactIDs. Previously we had such experience with `perforce-plugin` and `p4-plugin`: issue misreports, confused users, failed migration attempts. It will require careful documentation and maintenance of the bugtracker.

>  We are ok with just removing the old plugin from the marketplace to avoid confusion, as long as this new plugin is readily available and public.

Not sure that 1500 users  will share the concern. I would recommend to keep it around, especially if you consider some kind of maintenance for your customers. If no, maybe it makes sense to mark the plugin for adoption

CC @btk3000 just in case

slide:
Yes, if the original plugin is going to be deprecated, having them both exist for some time it fine. Is there an automatic upgrade path from the old plugin to the new one? If not, it might make sense to see if it is feasible.

akamen:
Those are good points.   Great, then we will co-exist for a little while, yes there does appear to be a good number of users however it is hard to determine how many are actual active users.  Is there a way to dissect those numbers  as the 1,500 appears to be downloads, right?

Agreed, the naming is always tough.  We want to eventually consolidate all the Jenkins plugins we have under one single umbrella.  Right now we have:  blackduck, blackduck-detect, coverity and soon synopsys-coverity.  Ideally all four will turn into one single presence:  synopsys.   Work in progress :D

oleg-nenashev:
>  Ideally all four will turn into one single presence: synopsys

Well, I hope all EDA plugins won't be put into the same plugin. Though `synopsys-software-integrity` would make sense

PR #831, opened 2018-09-03
Added initial version of permissions for analysis-model library.

uhafner:
# Description

The '[analysis-model](https://github.com/jenkinsci/analysis-model/)' project is a library to read static analysis reports into a Java object model. It is used by Jenkins' warnings plug-in.

### Always

- [ ] Add link to resolved HOSTING issue in description above

- [ ] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)
- [ ] Use the Jenkins community (LDAP) account name in the YAML file, not the GitHub account name
- [ ] [All newly added users have logged in to Artifactory at least once](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

### DIFF
--- a/permissions/component-analysis-model.yml
+++ b/permissions/component-analysis-model.yml
@@ -1,0 +1,7 @@
+---
+name: "analysis-model"
+github: "jenkinsci/analysis-model"
+paths:
+- "edu/hm/hafner/analysis-model"
+developers:
+- "drulli"


daniel-beck:
@uhafner There's a long history of libraries resulting in problems in Jenkins plugins due to diverging versions and/or the work required to update them. I recommend you consider wrapping this in a library plugin, and depend on the plugin instead of the lib directly, like we do for Async HTTP Client, Bouncycastle, and a few other libraries.

uhafner:
I'm not sure if I correctly understand the motivation for the problems. We have the same problems with plugins as well. It's a time consuming task to define a set of consistent dependencies of plugins (dependency hell and mvn enforcer).

If I use an additional library then I will have 3 modules 
warnings-plugin -> analysis-model-plugin -> analysis-model

The analysis-model-plugin is just empty with one dependency to analysis-model. Is this what you are proposing? (analysis-model should never depend on anything from Jenkins so that I can use it in a different application as well)

daniel-beck:
> The analysis-model-plugin is just empty with one dependency to analysis-model. Is this what you are proposing?

Yes. See https://plugins.jenkins.io/async-http-client https://plugins.jenkins.io/bouncycastle-api https://plugins.jenkins.io/apache-httpcomponents-client-4-api etc.

Rationale: https://jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/#using-library-wrapper-plugins

uhafner:
Ok, thanks. I'll add another repo and permission request then...

Should the permissions already work? I tried to release a RC version without success a couple of minutes ago.

daniel-beck:
@uhafner It should. I see RC2 was deployed so I assume this was a client-side problem.

uhafner:
Ok, thanks.

Followup: https://issues.jenkins-ci.org/browse/HOSTING-628

Seems that creating an empty plug-in is not supported by the api-plugin and the InjectedTests. I needed to add a dummy class to get everything working...

daniel-beck:
FWIW, something like `public class PluginImpl extends Plugin {}` should do it.

uhafner:
Yes I did something similar. Your tip is {{@Deprecated}} ;-)

PR #839, opened 2018-09-06
Add initial permissions for sysdig-secure plugin

alfred-landrum:
# Description

Adds initial permissions for the sysdig-secure plugin.
* https://issues.jenkins-ci.org/browse/HOSTING-617
* https://github.com/jenkinsci/sysdig-secure-plugin

### Always

### DIFF
--- a/permissions/plugin-sysdig-secure.yml
+++ b/permissions/plugin-sysdig-secure.yml
@@ -1,0 +1,7 @@
+---
+name: "sysdig-secure"
+github: "jenkinsci/sysdig-secure-plugin"
+paths:
+- "org/jenkins-ci/plugins/sysdig-secure"
+developers:
+- "alfredlandrum"
--- a/permissions/plugin-sysdig-secure.yml
+++ b/permissions/plugin-sysdig-secure.yml
@@ -5,3 +5,4 @@ paths:
 - "org/jenkins-ci/plugins/sysdig-secure"
 developers:
 - "alfredlandrum"
+- "sysdig"


permissions/plugin-sysdig-secure.yml
```diff
@@ -0,0 +1,7 @@
+---
+name: "sysdig-secure"
+github: "jenkinsci/sysdig-secure-plugin"
+paths:
+- "org/jenkins-ci/plugins/sysdig-secure"
+developers:
+- "alfredlandrum"
```
daniel-beck: We will consider this account to be the sole owner of the plugin. Since the Jira issue was filed as `sysdig`, it might make sense to add that as well, in case you change jobs.

alfred-landrum: Thank you, that does make sense. I've added that account, and we've logged in with that account to Artifactory.

daniel-beck:
@alfred-landrum What's going on in this plugin, why is it the same as https://github.com/jenkinsci/anchore-container-scanner-plugin ?

Please refrain from releasing, this will at least result in potential technical problems for users of both plugins.

alfred-landrum:
Hi @daniel-beck -  This is based on the Anchore plugin, with their knowledge & approval: Sysdig & Anchore have worked together so that Sysdig's Secure product offers container image scanning based on Anchore's anchore-engine: https://github.com/anchore/anchore-engine . The differences between anchore's plugin and this plugin are small today, but we expect to add more functionality over time.

@nurmi , would you mind confirming the above from Anchore's side?

@daniel-beck : We are new to Jenkins plugin development, so please let us know if there's something we need to do to our plugin code to avoid issues for any Jenkins users.

nurmi:
Hi @alfred-landrum @daniel-beck - we're working closely with sysdig and would like to support the effort to include modifications to the anchore plugin along with extended (sysdig specific) functionality in the future - we're more than happy to discuss the correct path forward to hit that objective!  @daniel-beck - can you make a recommendation on the right path forward, to help us better understand the steps we should jointly take, to start making the correct technical plan?

Best 
- Dan

daniel-beck:
> We are new to Jenkins plugin development, so please let us know if there's something we need to do to our plugin code to avoid issues for any Jenkins users.

Rename at least all packages. Otherwise having both plugins installed can result in classloading and persistence related problems.

PR #905, opened 2018-10-29
move org to com, add self

chook:


oleg-nenashev:
@chook it seems you removed the checklist without actually completing it

chook:
Hi Oleg - to be honest, Joe is asleep and I wanted to make sure the push i made to https://github.com/jenkinsci/overops-query-plugin actually reaches to the marketplace.
I may have went all in without actually understanding the fallouts.

I did try to do `mvn release:prepare release:perform -Dusername=chook -Dpassword=<pass>`
but i'm now failing on the scm-tag already existing. 

What I'm trying to do is to publish the new release. Let me know if I can without Joe's approval, and how. Thanks

chook:
@oleg-nenashev

batmat:
Hello, please do not remove the PR template as Oleg said. The failure is related to one of the checkboxes unfortunately blanked out.

Apart from this, unfortunately no, we cannot grant release permission without current maintainers' approval, we hope for obvious reasons.

batmat:
It looks like you already have commit access, which _could_ allow us accelerate the process.
Note that that kind of information you could provide with the description, and still does not prevent anyone from following the template. It's here to save everyone's time, especially reviewers' who are handling dozens of such requests, often on their free time.

chook:
fair enough. i'll look at the template

PR #936, opened 2018-11-21
Update tuleap-git-branch-source groupId to match Jenkins std

vaceletm:
# Description

Github: https://github.com/jenkinsci/tuleap-git-branch-source-plugin

### Always


daniel-beck:
@vaceletm Sorry about the delay. I've been busy with other work for the past few weeks, and it seems we have worse coverage here than I expected.

PR #938, opened 2018-11-26
Add @oleg-nenashev and @batmat to the list of Maven HPI Plugin maintainers

oleg-nenashev:
In the Java 11 maintenance team we will need to make some changes around HPI packaging, e.g. https://github.com/jenkinsci/maven-hpi-plugin/pull/75

CC @jglick @batmat

### DIFF
--- a/permissions/component-maven-hpi-plugin.yml
+++ b/permissions/component-maven-hpi-plugin.yml
@@ -11,4 +11,3 @@ developers:
 - "olamy"
 - "oleg_nenashev"
 - "batmat"
-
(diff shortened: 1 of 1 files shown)


permissions/component-maven-hpi-plugin.yml
```diff
@@ -9,3 +9,6 @@ developers:
 - "kohsuke"
 - "stephenconnolly"
 - "olamy"
+- "oleg_nenashev"
+- "batmat"
+
```
jglick: remove stray newline

PR #965, opened 2018-12-21
Use blocks for stage step

batmat:
Fixing the `Using the ‘stage’ step without a block argument is deprecated` warning.


batmat:
All seems fine, merging.
image

PR #972, opened 2018-12-28
Introduce local script to check before submitting

batmat:
_(PR on the tool itself, hence not following the PR template, as explained in the PR template :-))_

Many PRs fail on non existing usernames, etc. Let's make it easy for submitters to check before filing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,18 +41,11 @@ node('java') {
         }
 
         stage ('Run') {
-            def javaArgs = ' -DdefinitionsDir=$PWD/permissions' +
-                        ' -DartifactoryApiTempDir=$PWD/json' +
-                        ' -DartifactoryUserNamesJsonListUrl=https://reports.jenkins.io/artifactory-ldap-users-report.json' +
-                        ' -Djava.util.logging.SimpleFormatter.format="%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$s: %5$s%6$s%n"' +
-                        ' -jar target/repository-permissions-updater-*-bin/repository-permissions-updater-*.jar'
-
-
             if (dryRun) {
-                sh '${JAVA_HOME}/bin/java -DdryRun=true' + javaArgs
+                sh './update-permissions.sh'
             } else {
                 withCredentials([usernamePassword(credentialsId: 'artifactoryAdmin', passwordVariable: 'ARTIFACTORY_PASSWORD', usernameVariable: 'ARTIFACTORY_USERNAME')]) {
-                    sh '${JAVA_HOME}/bin/java ' + javaArgs
+                    sh './update-permissions.sh -f'
                 }
             }
         }
--- a/update-permissions.sh
+++ b/update-permissions.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -euo pipefail
+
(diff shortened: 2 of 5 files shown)


update-permissions.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
batmat: note: this script is in dry-run by default, given the main purpose is to enable submitters to do local check. Hence I do not want them to have to specify a switch for the nominal use case (and it's a no-brainer to do this specifically in the `Jenkinsfile`)

batmat:
https://github.com/jenkins-infra/repository-permissions-updater/pull/972/commits/6de0adbfd67fb49cb0abade8191cbe5661b0c192 pushed only for checking the tool still does the check in CI. Anyway, we will squash all as per the settings in this repo in the end.

batmat:
Failed as expected :+1: 
image

Jenkinsfile
```diff
@@ -41,18 +41,11 @@ node('java') {
...
-            def javaArgs = ' -DdefinitionsDir=$PWD/permissions' +
-                        ' -DartifactoryApiTempDir=$PWD/json' +
-                        ' -DartifactoryUserNamesJsonListUrl=https://reports.jenkins.io/artifactory-ldap-users-report.json' +
-                        ' -Djava.util.logging.SimpleFormatter.format="%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$s: %5$s%6$s%n"' +
-                        ' -jar target/repository-permissions-updater-*-bin/repository-permissions-updater-*.jar'
-
-
             if (dryRun) {
-                sh '${JAVA_HOME}/bin/java -DdryRun=true' + javaArgs
+                sh './update-permissions.sh'
```
batmat: Didn't bother making sure to reuse the path of the installed `JAVA_HOME`, it seems to still work. But can do if you think this is important.

daniel-beck:
I'm not really sure this is useful. It requires people to know how to build a Java Maven project, something we cannot assume is given.

Also, having to have logged in is part of the checklist. It's unclear how adding an additional, quasi redundant checklist entry is going to help -- except in the cases where users are too quick for the cached list of users to catch up. And in that case, perhaps we just need to be explicit about the need to wait 15 minutes, or until a user name appears in the JSON file, before filing a PR?

batmat:
@daniel-beck Well, I"ve seen roughly 1 PR out of 2 lately failing because the username was wrong, and not because they didn't log in or so. So having a way to locally check things look good before submitting would remove _some_ of the broken PRs from the ground up.

Also, if you're worried by people having to build a Java project, which makes some sense (though generally, they should have it configured, given they are generally people who are writing a Jenkins plugin, using Java & Maven), I can wrap that check tool in a Docker container so no local setup is needed (but `Docker`).

daniel-beck:
TBH I would prefer better reporting of PR failures to a (manual) prerequisite check. Perhaps just needs to change error presentation to get rid of the stack trace, change error message to refer to the JSON file URL, or mention a caching delay, or…

PR builds _are_ the automatic prerequisite check, after all 😃

batmat:
Could be, but they also generate notifications. And as I'm watching the repo, it's a waste of time when people keep submitting malformed PRs. :disappointed:

PR #998, opened 2019-01-17
Ignore json/ directory

batmat:
Patch to the tool itself, not permissions related.


batmat:
Moving forward to cleanup the PR queue, and also obviously given the low level of risk this change has. Worst case, I'll revert.

PR #999, opened 2019-01-17
Add jaxb permission file

batmat:
# Description

This PR is a special one: it creates a permission file for https://github.com/jenkinsci/jaxb-plugin that got created directly by @kohsuke in the context of Java 11 testing.
I already got commit access to that repository.

My only goal here is to enable Incremental releases, which I believe is requiring the `github` field. That is why there's no developers specified to be allowed to release.

Thanks!

### Always

- ~Add link to resolved HOSTING issue in description above~ N/A

- [x] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

cc @jenkins-infra/java11-support

### DIFF
--- a/permissions/plugin-jaxb.yml
+++ b/permissions/plugin-jaxb.yml
@@ -3,4 +3,7 @@ name: "jaxb"
 github: "jenkinsci/jaxb-plugin"
 paths:
 - "io/jenkins/plugins/jaxb"
-developers: []
+developers:
+- "kohsuke"
+- "oleg_nenashev"
+- "batmat"
--- a/permissions/plugin-jaxb.yml
+++ b/permissions/plugin-jaxb.yml
@@ -3,6 +3,4 @@ name: "jaxb"
 github: "jenkinsci/jaxb-plugin"
 paths:
 - "io/jenkins/plugins/jaxb"
-developers:
-- "batmat"
-- "oleg_nenashev"
+developers: []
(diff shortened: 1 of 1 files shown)


batmat:
Cf. https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fjaxb-plugin/detail/master/2/pipeline, goal is to have the Incrementals deployment be successful.

Currently failing as:

image

oleg-nenashev:
Formally it needs authorization from @kohsuke 
But I have no strong opinion in this case

batmat:
@oleg-nenashev I think we do not absolutely need @kohsuke approval in this very case, though for sure it would be better/cleaner, for at least the following reasons:
* I think I remember a dev list thread, but can't find it just now, where Kohsuke already stated that anyone should feel free to take over maintainership because he didn't consider himself anymore an active maintainer of any plugin
* that plugin is still in a very early stage, has never been released yet, and was created solely to see how Jenkins could support JAXB once on Java 11. 
   * As you and I are members of the @jenkins-infra/java11-support team, taking over this does seem to make sense
* Given Kohsuke has not committed in that repo since June 2018, I suppose anyway we could on top of everything consider it abandoned.

Again, I agree it would be cleaner to have his approval confirmed. But I think we could move forward because I suppose he would trust us to not do something evil in such case.

oleg-nenashev:
Thanks @kohsuke !

PR #1010, opened 2019-01-25
Update plugin-benchmark-evaluator.yml

Lehmann-Fabian:
# Description

https://github.com/jenkinsci/benchmark-evaluator-plugin

New reference to "org/jenkins-ci/plugins/benchmarking-evaluator" to make it visible in https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/ Currently the plugin is not available in the Jenkins market. 

### Always

### DIFF
--- a/permissions/plugin-benchmark-evaluator.yml
+++ b/permissions/plugin-benchmark-evaluator.yml
@@ -3,6 +3,7 @@ name: "benchmark-evaluator"
 github: "jenkinsci/benchmark-evaluator-plugin"
 paths:
 - "io/jenkins/plugins/benchmark-evaluator"
+- "org/jenkins-ci/plugins/benchmark-evaluator"
 developers:
 - "lehmann_fabian"
 - "martingrambow"


batmat:
What are you trying to achieve? Please, in general, that is a good practice to explain your intent up front so we all save time :crying_cat_face:

batmat:
Also, 1) multiple paths are not supported 2) this doesn't match your plugin's GAV, and you removed parts of the PR template that would have reminded that and 3) I suspect you're filing this because you don't see your plugin in the update center yet. 

But it is released: https://repo.jenkins-ci.org/releases/io/jenkins/plugins/benchmark-evaluator/1.0.0/

There's an ongoing issue with the Update Center. Please in general be patient, and expect your releases to be seen by users after ~12 hours at least. Avoid asking questions before so we all save a bit time :). Thanks!

Lehmann-Fabian:
@batmat Thank you very much for your answer. That is exactly what I meant. I had oriented myself to the file "plugin-analysis-model-api.yml", which also has two paths.

PR #1048, opened 2019-03-05
Grant @oleg-nenashev and @olivergondza permissions to release the Javadoc plugin

oleg-nenashev:
CC @jglick 

Subj: https://github.com/jenkinsci/javadoc-plugin/pull/10

### DIFF
--- a/permissions/plugin-javadoc.yml
+++ b/permissions/plugin-javadoc.yml
@@ -5,3 +5,5 @@ paths:
 - "org/jenkins-ci/plugins/javadoc"
 developers:
 - "jglick"
+- "oleg_nenashev"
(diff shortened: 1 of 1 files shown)


jglick:
and CC @olivergondza

oleg-nenashev:
ugh. How did I manage to create it from the master fork?
Should I recreate it @batmat ?

batmat:
I don't feel that strongly about that, was trying to retrigger, but I would suspect the PR build is configured to only watch out forks. Which might be why the PR build doesn't kick in (but I've no access to the config).

oleg-nenashev:
created the new one

PR #1055, opened 2019-03-08
Grant @oleg-nenashev permissions to release the Unique ID plugin

oleg-nenashev:
It was approved by @stephenc in https://github.com/jenkinsci/unique-id-plugin/pull/12#issuecomment-470873306

### DIFF
--- a/permissions/plugin-unique-id.yml
+++ b/permissions/plugin-unique-id.yml
@@ -8,3 +8,4 @@ developers:
 - "oleg_nenashev"
 - "recampbell"
 - "stephenconnolly"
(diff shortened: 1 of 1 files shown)


permissions/plugin-unique-id.yml
```diff
@@ -7,4 +7,5 @@ developers:
 - "andresrc"
 - "oleg_nenashev"
 - "recampbell"
-- "stephenconnolly"
\ No newline at end of file
+- "stephenconnolly"
+- "oleg_nenashev"
```
daniel-beck, line 11: wat

PR #1065, opened 2019-03-20
Add security contacts metadata

daniel-beck:
So far we tracked this internally, but it cannot hurt adding it to the metadata here, and giving maintainers an easy way to define this themselves.

### DIFF
--- a/permissions/plugin-github.yml
+++ b/permissions/plugin-github.yml
@@ -7,3 +7,6 @@ developers:
 - "integer"
 - "lanwen"
 - "oleg_nenashev"
+security:
+  contacts:
+    jira: "integer"
--- a/permissions/plugin-maven-plugin.yml
+++ b/permissions/plugin-maven-plugin.yml
@@ -11,3 +11,6 @@ developers:
 - "olamy"
 - "olivergondza"
 - "aheritier"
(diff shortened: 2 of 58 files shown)


permissions/plugin-github.yml
```diff
@@ -7,3 +7,6 @@ developers:
 - "integer"
 - "lanwen"
 - "oleg_nenashev"
+security:
+  contacts:
+    jira: "integer"
```
daniel-beck: FYI @KostyaSha

daniel-beck:
@halkeye Could you review this to see whether I included everything in your security group's domain?

permissions/plugin-maven-plugin.yml
```diff
@@ -11,3 +11,6 @@ developers:
 - "olamy"
 - "olivergondza"
 - "aheritier"
+security:
```
daniel-beck: @olamy @aheritier FYI

halkeye:
> @halkeye Could you review this to see whether I included everything in your security group's domain?

I think that's good. Nothing stands out as missing. Officially I think we also maintain the jira-plugin (not the blueocean integration) but i don't think that makes sense.

PR #1076, opened 2019-03-29
Add security contact for several plugins

daniel-beck:
@csanchez @Vlatombe As maintainers, please approve or contact @richbg.

FYI @maxlaverse @maselvaraj this is seeking to define who would be informed about security vulnerabilities in these plugins. In the past I've chosen a maintainer at random. Docs: https://github.com/jenkins-infra/repository-permissions-updater#managing-security-process

### DIFF
--- a/permissions/plugin-artifact-manager-s3.yml
+++ b/permissions/plugin-artifact-manager-s3.yml
@@ -7,3 +7,6 @@ developers:
 - "jglick"
 - "csanchez"
 - "oleg_nenashev"
+security:
+  contacts:
+    jira: "cloud_platform_v2"
--- a/permissions/plugin-kubernetes-credentials.yml
+++ b/permissions/plugin-kubernetes-credentials.yml
@@ -7,3 +7,6 @@ developers:
 - "max_laverse"
 - "csanchez"
 - "sirstrahd"
+security:
+  contacts:
+    jira: "cloud_platform_v2"
--- a/permissions/plugin-mesos.yml
+++ b/permissions/plugin-mesos.yml
@@ -7,3 +7,6 @@ developers:
 - "maselvaraj"
 - "vinodkone"
 - "vlatombe"
+security:
+  contacts:
+    jira: "cloud_platform_v2"


maselvaraj:
Sure. Thanks @daniel-beck  👍

daniel-beck:
Premature, lacking approval from artifact-manager-s3 and kubernetes-credentials maintainers.

slide:
My apologies, I misunderstood the approvals that were waiting.

maxlaverse:
Sorry, I overlooked the first GitHub notification.

The `kubernetes-credentials` plugin is backed by individuals and doesn't have a dedicated security team but I'd like to be informed if there are security vulnerabilities in it.

Who is represented by `cloud_platform_v2` @slide ? It's not mentioned in the link you pasted and apparently it's not used for any other plugin than the 3 in this pull-request.

daniel-beck:
@maxlaverse It's an account backed by a team at CloudBees (I believe one that Carlos and Vincent are on, hence the choice for these plugins). We started doing this quite a bit recently -- e.g. most of the Pipeline and Blue Ocean plugins are covered similarly. The idea here is that doing so makes it much less likely that issues fall between the cracks and are forgotten.

> I'd like to be informed if there are security vulnerabilities in it.

From a Jenkins security team perspective, we can only add one assignee to Jira issues, and doing further out of bands communication is simply too much overhead. So I would ask that you align with your co-maintainers here. Maybe it's good enough to leave the security contact as defined here, and they keep you informed out of band -- something we recommend anyway to prevent regular releases from interfering with security update preparation (something that's happened [with Mesos Plugin before](https://github.com/jenkinsci/mesos-plugin/commits/master) because only Vincent was aware this was happening). 

If you'd prefer to revert this change in the mean time, please file a PR to revert your plugin's part of this change and we'll go back to the default of "choose one of the maintainers as contact for security issues" until a security contact is defined.

maxlaverse:
Sure, I understand.
I'm perfectly fine with having that team from CloudBees notified first. It's a pretty small plugin with only a few lines of code and the initial implementation was done by another CloudBees employee anyway.
Thanks for the explanation @daniel-beck

permissions/plugin-artifact-manager-s3.yml
```diff
@@ -7,3 +7,6 @@ developers:
 - "jglick"
 - "csanchez"
 - "oleg_nenashev"
+security:
+  contacts:
+    jira: "cloud_platform_v2"
```
jglick: Not sure exactly what this is, but if it goes to at least @carlossg then that is fine.

richbg: @jglick carlossg has been added to the distro list

PR #1170, opened 2019-06-11
Replace dwnusbaum with mramonleon in metrics plugin

dwnusbaum:
# Description

Replaces myself with @MRamonLeon as a maintainer of https://github.com/jenkinsci/metrics-plugin. I originally became a maintainer of this plugin as it was related to my work, but I am working on a different team now (And fwiw I didn't do much with the plugin in the first place, I just filed one PR and merged one from someone else). Approved by @dwnusbaum  as a current maintainer of the plugin, CC @alecharp if approval from another maintainer is needed.

Keep in mind that @stephenc requests that any changes to the public API of the plugin be reviewed by him first.

# Submitter checklist for changing permissions

### Always

- [x] Add link to plugin/component Git repository in description above

### When adding new uploaders (this includes newly created permissions files)

- [x] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)
- [x] Use the Jenkins community (LDAP) account name in the YAML file, not the GitHub account name
- [x] [All newly added users have logged in to Artifactory at least once](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

PR #1172, opened 2019-06-12
Change case of mramonleon

MRamonLeon:
# Description

Just a change in the case of my user name. I think it matters because I haven't any rights on the previous GH repositories I was included.


slide:
These files are not for github access, they are for access to release to the maven repository.

batmat:
I think this was solved with https://github.com/jenkins-infra/repository-permissions-updater/pull/1170#issuecomment-501219737

PR #1184, opened 2019-06-19
missing quote

cryptomoovin:
sorry to bother, thanks for helping. related to #1183


slide:
I should check more carefully on these...

PR #1198, opened 2019-06-27
Grant @oleg-nenashev and @raul-arabaolaza permissions to release PCT

oleg-nenashev:
Subj. Just recovering the release flow for PCT which becomes more and more important to the organization. Approving it as a PCT maintainer


oleg-nenashev:
Uh oh: Waiting for next available executor on ‘java’

PR #1211, opened 2019-07-06
Create plugin-graphql-server.yml

halkeye:
# Description

https://issues.jenkins-ci.org/browse/HOSTING-784
https://github.com/jenkinsci/graphql-server-plugin
No previous maintainers

### Always

- [X] Add link to plugin/component Git repository in description above

- [X] Add link to resolved HOSTING issue in description above

- [X] Make sure the file is created in `permissions/` directory
- [X] `artifactId` (pom.xml) is used for `name` (permissions YAML file).
- [X] [`groupId` / `artifactId` (pom.xml) are correctly represented in `path` (permissions YAML file)](https://github.com/jenkins-infra/repository-permissions-updater/#managing-permissions)
- [X] Check that the file is named `plugin-${artifactId}.yml` for plugins

### DIFF
--- a/permissions/plugin-graphql-server.yml
+++ b/permissions/plugin-graphql-server.yml
@@ -4,4 +4,4 @@ github: "jenkinsci/graphql-plugin"
 paths:
 - "org/jenkins-ci/plugins/graphql-plugin"
 developers:
-- "halkeye
+- "halkeye"
--- a/permissions/plugin-graphql-server.yml
+++ b/permissions/plugin-graphql-server.yml
@@ -1,0 +1,7 @@
+---
+name: "graphql-plugin"
+github: "jenkinsci/graphql-plugin"
+paths:
+- "org/jenkins-ci/plugins/graphql-plugin"
+developers:
+- "halkeye
--- a/permissions/plugin-graphql-server.yml
+++ b/permissions/plugin-graphql-server.yml
@@ -2,6 +2,6 @@
 name: "graphql-plugin"
 github: "jenkinsci/graphql-plugin"
 paths:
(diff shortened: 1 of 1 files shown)


permissions/plugin-graphql-server.yml
```diff
@@ -0,0 +1,7 @@
+---
+name: "graphql-plugin"
+github: "jenkinsci/graphql-plugin"
+paths:
+- "org/jenkins-ci/plugins/graphql-plugin"
```
slide: This path is incorrect, it should match the groupId/artifactId from your pom.xml (https://github.com/jenkinsci/graphql-server-plugin/blob/master/pom.xml#L9). So, you would need "io/jenkins/plugins/graphql-server"

halkeye: thanks, skimmed it too fast.

slide:
Looks like the name and GitHub Peyton are incorrect too. The plugin repo is  graphql-server-plugin, and the name is graphql-server.

halkeye:
thanks for catching that too
i fixed one earlier, I think they line up now.

permissions/plugin-graphql-server.yml
```diff
@@ -0,0 +1,7 @@
+---
+name: "graphql-plugin"
```
slide: I think the name still needs to be graphql-server (plugin name in pom?)

halkeye:
@slide your absolutely right. I've been doing most of the fixes to the file on mobile and not giving it enough attention.

I hope that's the last fix I do on mobile.

I'll be back on a computer in a couple hours and look more carefully.

PR #1278, opened 2019-09-16
Grant @oleg-nenashev permissions in the Extended Read Plugin

oleg-nenashev:
I would like to release https://github.com/jenkinsci/extended-read-permission-plugin/pull/4 and https://github.com/jenkinsci/extended-read-permission-plugin/pull/3 so that we have a refreshed of the plugin for instances using newer core versions.

I am not sure whether it is fine to bother @batmat w.r.t releases, so raising this pull request just in case he wants to share ownership


slide:
@batmat Any issues with this?

PR #1281, opened 2019-09-20
Removing developer from project

chris-overend:
https://github.com/jenkinsci/vectorcast-coverage-plugin

This developer is no longer with the company

### DIFF
--- a/permissions/plugin-vectorcast-coverage.yml
+++ b/permissions/plugin-vectorcast-coverage.yml
@@ -6,4 +6,3 @@ paths:
 developers:
 - "rmkeeble"
 - "chris_overend"
-- "matthew_queenan"


slide:
@oleg-nenashev When someone leaves a company, do we need to wait for confirmation?

oleg-nenashev:
Yes. There is no notion of company ownership ATM. I started drafting JEP,
but have not got to it yet

daniel-beck:
The requester here is a maintainer, therefore this request can be implemented – we'll just have to wait a bit for the affected user to comment, to prevent us assisting in some personal conflict between maintainers. This is different from general cleanup where e.g. I remove someone from components I do _not_ maintain.

slide:
It's been over a month here, I think we can move forward? @daniel-beck @oleg-nenashev any issues?

daniel-beck:
Nope, this was fine after a week timeout already IMO.

PR #1286, opened 2019-09-26
[JENKINS-58770] Be able to release new artifactId’s

jglick:
# Description

Needed to make https://github.com/jenkinsci/bom/pull/107 releasable.

### DIFF
--- a/permissions/pom-bom.yml
+++ b/permissions/pom-bom.yml
@@ -2,6 +2,10 @@
 name: "bom"
 github: "jenkinsci/bom"
 paths:
-- "io/jenkins/tools/bom/bom"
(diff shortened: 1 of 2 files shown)


permissions/pom-bom.yml
```diff
@@ -2,6 +2,10 @@
 name: "bom"
 github: "jenkinsci/bom"
 paths:
-- "io/jenkins/tools/bom/bom"
+- "io/jenkins/tools/bom/bom-2.190.x"
```
casz: Could be reduced to so you do not have to repeat requesting permission for new Jenkins versions.
```suggestion
- "io/jenkins/tools/bom"
```

PR #1297, opened 2019-10-09
Create plugin-pipeline-cps-http.yml

julienduchesne:
Just got this plugin hosted
https://github.com/jenkinsci/pipeline-cps-http-plugin
https://issues.jenkins-ci.org/browse/HOSTING-837


slide:
Your path is incorrect. It should be your groupId/artifactId is as defined in your pom.xml. in your case, the groupId is io.jenkins.plugins, so your path would be io/jenkins/plugins/YOURARTIFACTID

julienduchesne:
Oops. Here you go. Thank you

PR #1344, opened 2019-11-13
Create CODEOWNERS to automatically request reviews from the Hosting team

oleg-nenashev:
Subj. I created a new team so that we can load-balance the hosting requests and automatically request reviews on new pull requests.

CC @daniel-beck @batmat @slide who are in the team now

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @jenkins-infra/hosting-team
+* @jenkins-infra/hosting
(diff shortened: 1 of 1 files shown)


slide:
Creating a branch in the same repo and then creating a PR from that branch causes ci.j.io some issue, the PR itself is never built, only the branch.

oleg-nenashev:
> Isn't it a bit redundant to specify 'team' in the team name?

Can rename it

PR #1384, opened 2020-01-02
Add security contact

diogo-duarte-rg:
Add security contact in yaml file for https://github.com/jenkinsci/redgate-sql-ci-plugin


batmat:
Note to self, we really need the information to mapping gh user to JIRA/infra one... I assume the submitter is the one declared already.

batmat:
>  IIRC we have the mapping in the account app. 

Yes, I _think_ Olivier said this is quite incomplete currently, give it's not required.

> I do not know how to retrieve this data tho

That was mostly my point ;-)

PR #1393, opened 2020-01-07
Add request to @ the user's being added

slide:
# Description

The LDAP username for the user as used in the .yml file may be different than the GitHub username which is needed to add GitHub permissions. Request that the users being added in the yml file are @ mentioned as well so we get their GitHub usernames.


oleg-nenashev:
CI does not work well for PRs from the same repo

PR #1422, opened 2020-02-10
remove me as maintainer

imod:
I only have limited time available for plugin maintenance and I would like to spend it on the plugins I use myself


oleg-nenashev:
Thanks for your contributions @imod !

PR #1510, opened 2020-04-26
Modernize tool

daniel-beck:
* Use SHA-256 instead of MD5 to shorten file names (doesn't matter but good hygiene)
* [INFRA-2594](https://issues.jenkins-ci.org/browse/INFRA-2594): Add support for new Gradle versions
* Remove stray top-level permissions file
* General dependency modernization, subsumes #1447

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci</groupId>
         <artifactId>jenkins</artifactId>
-        <version>1.36</version>
+        <version>1.55</version>
     </parent>
 
     <groupId>io.jenkins.infra</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -66,28 +66,35 @@
...
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
-            <version>2.8.11</version>
+            <version>${jackson.version}</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.8.11.3</version>
+            <version>${jackson.version}</version>
```
slide, line 87: This version doesn't necessarily follow the versions of the other two dependencies from what I have seen.

daniel-beck: Can be changed if they deviate again. That said, I think that was mostly for the security issues that are obsolete with 2.10.x

daniel-beck:
Someone will complain if permissions are gone 🤷

PR #1551, opened 2020-05-28
More meaningful error when checking YAML names

ahmed-musallam:
Better error message when checking if file name ends with `.yml`


ahmed-musallam:
@slide looks like this is good to be merged. Thanks for all your help!

PR #1576, opened 2020-06-16
Fix the sideex permission file

oleg-nenashev:
It is a follow-up to the issue reported by the SideEx team in private. CC @kiam123 .
Release attempts fail due to the wrong path in the file.

### DIFF
--- a/permissions/plugin-sideex.yml
+++ b/permissions/plugin-sideex.yml
@@ -2,6 +2,6 @@
 name: "sideex"
 github: "jenkinsci/sideex-plugin"
 paths:
-- "io/jenkinsi/plugins/sideex"
+- "io/jenkins/plugins/sideex"
 developers:
 - "sideexteam"


kiam123:
Hi, we are still having the same error after your commit, is it because the path name should be "jenkins-ci" instead of "jenkins"?

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.8.2:deploy (default-deploy) on project sideex: Failed to deploy artifacts: Could not transfer artifact io.jenkins.plugins:sideex:hpi:1.0 from/to maven.jenkins-ci.org (https://repo.jenkins-ci.org/releases/): Access denied to: https://repo.jenkins-ci.org/releases/io/jenkins/plugins/sideex/1.0/sideex-1.0.hpi, ReasonPhrase: Forbidden. -> [Help 1]

rbywater:
@kiam123 Note that the PR has not been merged yet so the changes will not have taken effect.

kiam123:
@rbywater ok, thank you for your reply

oleg-nenashev:
It should be good to go in 15min or so

PR #1779, opened 2020-12-15
[JEP-229] Fix Jenkinsfile to provide credentials; enable CD for log-cli

daniel-beck:
Follow up on https://github.com/jenkins-infra/repository-permissions-updater/pull/1747 with some unfinished WIP I had forgotten about 😢 

Anyway, this seems to work now, tested by switching the job on trusted.ci over to this origin branch (and back again).

@jglick clearly is fine with `log-cli` being CD enabled, so that's why I set that up here already.

### DIFF
--- a/permissions/plugin-log-cli.yml
+++ b/permissions/plugin-log-cli.yml
@@ -1,6 +1,8 @@
 ---
 name: "log-cli"
 github: "jenkinsci/log-cli-plugin"
+cd:
+  enabled: true
 paths:
 - "org/jenkins-ci/plugins/log-cli"
 developers:
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -70,7 +70,10 @@ node('java') {
                         name: 'Validation',
                         title: 'All checks passed'
             } else {
-                withCredentials([usernamePassword(credentialsId: 'artifactoryAdmin', passwordVariable: 'ARTIFACTORY_PASSWORD', usernameVariable: 'ARTIFACTORY_USERNAME')]) {
+                withCredentials([
+                        usernamePassword(credentialsId: 'artifactoryAdmin', passwordVariable: 'ARTIFACTORY_PASSWORD', usernameVariable: 'ARTIFACTORY_USERNAME'),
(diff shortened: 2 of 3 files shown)


permissions/plugin-log-cli.yml
```diff
@@ -1,6 +1,8 @@
 ---
 name: "log-cli"
 github: "jenkinsci/log-cli-plugin"
+cd:
+  enabled: true
```
jglick, line 5: :+1:

jglick (approved):

daniel-beck:
FTR this is deliberately still undocumented, I don't think we're quite ready to declare this JEP active.

jglick:
> I don't think we're quite ready to declare this JEP active.

Yes, can still be in **Draft** status until some more pieces are implemented and tested. @oleg-nenashev I would still really like to get https://github.com/jenkinsci/jep/pull/332 merged though, so that the published edition more closely reflects the various changes that are being made.

timja (approved):

jglick:
I see `MAVEN_USERNAME` + `MAVEN_TOKEN` in https://github.com/jenkinsci/log-cli-plugin/settings/secrets/actions, great…

jglick:
https://github.com/jglick/jenkins-maven-cd-action/issues/2 accordingly.

timja:
This ready for use in other plugins @jglick?

jglick:
@timja yes if you are ready to live with very experimental stuff; it is up and running in `log-cli-plugin` anyway. I suspect many maintainers will want https://github.com/jglick/jenkins-maven-cd-action/issues/1, and https://github.com/jglick/verify-ci-status-action/issues/1 is likely to become a priority if changes proposed on jenkins-dev go ahead, and I plan to request `HOSTING` to get these tools into @jenkinsci or @jenkins-infra. For the near term I would recommend use on relatively minor plugins to minimize the impact of any grave problems that might arise.

I actually think that non-plugin components (`bom`, `jenkins-test-harness`) would be good candidates too—there is nothing about this system that is limited to `*-plugin` that I know of. Just need to be incrementalified, deployed to Artifactory, and managed via RPU, typically via `org.jenkins-ci:jenkins`. (`archetypes` and other components deployed to OSSRH, typically via `org.kohsuke:pom`, are not eligible. `plugin-pom` is also probably not incrementalifiable for technical reasons.)

timja:
Experimental is fine with me :)

PR #1783, opened 2020-12-17
Fix access-modifier deployment hopefully

timja:
See https://github.com/jenkinsci/lib-access-modifier/issues/41

### DIFF
--- a/permissions/component-kohsuke-maven-plugins.yml
+++ b/permissions/component-kohsuke-maven-plugins.yml
@@ -1,0 +1,12 @@
+---
+# this allows publishing one of 'kohsuke's maven plugins e.g. 'access-modifier'
+name: "kohsuke-maven-plugins" # not actually a single artifact
+github: "jenkinsci/lib-access-modifier" # not really true but will it break without?
+paths:
+- "org/kohsuke"
+developers:
+- "timja"
+- "oleg_nenashev"
+- "jglick"
+- "batmat"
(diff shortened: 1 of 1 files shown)


permissions/component-kohsuke-maven-plugins.yml
```diff
@@ -0,0 +1,12 @@
+---
+# this allows publishing one of 'kohsuke's maven plugins e.g. 'access-modifier'
+name: "kohsuke-maven-plugins" # not actually a single artifact
```
daniel-beck, line 3:
```suggestion
name: "dummy-unused-artifact-name"
```

permissions/component-kohsuke-maven-plugins.yml
```diff
@@ -0,0 +1,12 @@
...
+github: "jenkinsci/lib-access-modifier" # not really true but will it break without?
```
daniel-beck, line 4: Not all files have this, safe to leave out.

```suggestion
```

permissions/component-kohsuke-maven-plugins.yml
```diff
@@ -0,0 +1,12 @@
+---
+# this allows publishing one of 'kohsuke's maven plugins e.g. 'access-modifier'
```
daniel-beck, line 2:
```suggestion
# Workaround for publishing Maven plugins
```

PR #1793, opened 2020-12-22
Add issue tracker information to most of the plugins

halkeye:
See mailing list - https://groups.google.com/g/jenkinsci-dev/c/LnvX3mX5Gek

Add specific issue tracking information to all the plugins.


permissions/plugin-GatekeeperPlugin.yml
```diff
@@ -5,4 +5,5 @@ paths:
 - "org/paylogic/GatekeeperPlugin"
 developers:
 - "bubenkoff"
-- "maikelwever"
\ No newline at end of file
+- "maikelwever"
+issues: "github:jenkinsci/gatekeeper-plugin"
```
timja, line 9: do we need to specify the github repo for the issue tracker or should we be pulling that from the `github` field? as a backup we could have just github and allow overriding it.

I'm concerned we're duplicating information when we don't need to

halkeye: > I'm concerned we're duplicating information when we don't need to

I never want to guess. 

I also don't see the problem with duplicating the information, is there a byte limit thing your concerned about?

timja: it's not guessing though, a github issue tracker is tied to the repo.

I'm not aware of any plugin that tracks issues in another github issue tracker (not to say we can't support it)

> I also don't see the problem with duplicating the information, is there a byte limit thing your concerned about?

no it's just people having to duplicate information, why do it?

timja:
If you can put the issue tracker not at the bottom this PR is less likely to conflict

slide:
I'm no yaml expert by any means, but to make it easier to use this data, couldn't the type by split out and then only provide an extra piece of information if it is different from either the github field (for github issues) and name field (for jira)?

This would assume that the repo is the same as the `github` field
```
github: "jenkinsci/awesome-plugin"
issues: 
  type: "github"
  
```  

This explicitly specifies the repo
```
github: "jenkinsci/another-awesome-plugin"
issues:
  type: "github"
  repo: "jenkinsci/some-weird-plugin"
```

This assumes a Jira component named `amazing-plugin`
```
name: amazing-plugin
issues:
  type: "jira"
```

This specifies a different component in Jira
```
name: another-amazing-plugin
issues:
  type: "jira"
  component: "really-amazing-plugin"
```

halkeye:
> If you can put the issue tracker not at the bottom this PR is less likely to conflict

I'm just string concatinating, so its easier to add to the end. Its also super easy to rebuild the PR so i'm not going to worry about it.
```
git fetch orign
git reset origin/master
git checkout permissions/
groovy update-issues.groovy
git add permissions/
git commit -m "Add issue trackers to all plugins"
git push -f
```

> I'm no yaml expert by any means, but to make it easier to use this data, couldn't the type by split out and then only provide an extra piece of information if it is different from either the github field (for github issues) and name field (for jira)?

I'm still going to keep it explicit because as a consumer its way easier for me. But totally up for splitting it up. Going to keep it the same names though, so type/value instead of type/github and type/component

timja:
> I'm still going to keep it explicit because as a consumer its way easier for me. But totally up for splitting it up. Going to keep it the same names though, so type/value instead of type/github and type/component

can you post an example of what you mean?

halkeye:
> can you post an example of what you mean?

https://github.com/jenkins-infra/plugin-site-api/blob/master/src/main/java/io/jenkins/plugins/services/impl/HttpClientJiraIssues.java#L49

Some plugins names are uppercase, some are lower case, some have plugin, some don't.

By being explicit, i just have to read the data once, match it up once, and don't have to worry about things

We didn't have to have github: key in the yaml, cause it often matches up, but being explicit means you don't need to fuss with it, just use the data.

timja:
better but I still think we should infer the issue tracker from the github field,
if you really want to have it explicit I think you could use a yaml anchor and then at resolution time you still get the field?

halkeye:
>  if you really want to have it explicit I think you could use a yaml anchor and then at resolution time you still get the field?

Thats way more yaml than I know, It involves & right?

I can take a stab at it later, personally I don't think its worth the effort.

I've updated https://gist.github.com/halkeye/7f30c8385f28c40f99f66967ab74b42b if you have the time to poke at it.

(also need to figure out if its worth adding that script to this PR)

timja:
Ok I don't think that the yaml anchor bit is worth it,

I'm still not sold on the duplication of information.

Is it worth exploring what the code would look like in other components?
update-center / plugin-site?

I assume update-center is the next location so that both plugin-site and jenkins core can include the information?

PR #1806, opened 2021-01-17
Add issue tracker support, add output as issues.index.json

daniel-beck:
This is mostly the tool support side of things, with two files changed thrown in to demonstrate the effect (@timja to check whether I got the metadata correct for JCasC).

I switched from Jackson to SnakeYAML because the former does not support YAML references. Those make it less painful to deal with the "duplicated" GitHub URL without custom syntax, see README for explicit example.

Previously discussed in:

- https://gist.github.com/daniel-beck/a1a3fe34e2e5f27d56d64fee1de11c0e
- https://groups.google.com/g/jenkinsci-dev/c/LnvX3mX5Gek

Output created by the "sample" metadata changes included here:

```
$ cat json/issues.index.json 
{
    "configuration-as-code": [
        {
            "type": "github",
            "reference": "jenkinsci/configuration-as-code-plugin",
            "viewUrl": "https://github.com/jenkinsci/configuration-as-code-plugin/issues",
            "reportUrl": "https://github.com/jenkinsci/configuration-as-code-plugin/issues/new/choose"
        },
        {
            "type": "jira",
            "reference": "23170",
            "viewUrl": "https://issues.jenkins.io/issues/?jql=component=23170"
        }
    ],
    "matrix-auth": [
        {
            "type": "jira",
            "reference": "matrix-auth-plugin",
            "viewUrl": "https://issues.jenkins.io/issues/?jql=component=18131",
            "reportUrl": "https://issues.jenkins.io/secure/CreateIssueDetails!init.jspa?pid=10172&issuetype=1&components=18131"
        }
    ]
}
```

Both the configuration (YAML) and downstream consumption (JSON) formats are specified in the README, please review these definitions to ensure nothing is missing or ambiguous.

This file would be created in https://ci.jenkins.io/job/Infra/job/repository-permissions-updater/job/master/lastSuccessfulBuild/artifact/json/ but we probably want to add it to `reports.jenkins.io` from `trusted.ci` `Jenkinsfile` execution as a final canonical location.

FWIW there are very few conditions that cause failures here: E.g. if a Jira component is specified and that gets renamed, this tool shouldn't fail IMO.

### DIFF
--- a/permissions/plugin-configuration-as-code.yml
+++ b/permissions/plugin-configuration-as-code.yml
@@ -1,6 +1,6 @@
 ---
 name: "configuration-as-code"
-github: "jenkinsci/configuration-as-code-plugin"
+github: &GH jenkinsci/configuration-as-code-plugin
 paths:
 - "org/jenkins-ci/plugins/configuration-as-code"
 - "io/jenkins/configuration-as-code"
@@ -12,3 +12,7 @@ developers:
 security:
   contacts:
     jira: "jetersen"
+issues:
+  - github: *GH
+  - jira: 23170 # configuration-as-code-plugin
+    report: no
--- a/src/main/groovy/io/jenkins/infra/repository_permissions_updater/JiraAPI.groovy
+++ b/src/main/groovy/io/jenkins/infra/repository_permissions_updater/JiraAPI.groovy
@@ -1,0 +1,66 @@
+package io.jenkins.infra.repository_permissions_updater
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
+import groovy.json.JsonSlurper
+
+import java.util.logging.Level
+import java.util.logging.Logger
+
+@SuppressFBWarnings("LI_LAZY_INIT_STATIC") // Something related to Groovy
+abstract class JiraAPI implements Definition.IssueTracker.JiraComponentSource {
+    public static final Logger LOGGER = Logger.getLogger(JiraAPI.class.getName())
+    /**
+     * URL to Jira
+     */
+    private static final String JIRA_URL = System.getProperty('jiraUrl', 'https://issues.jenkins.io')
+    /**
+     * URL to Jira components API
+     */
+    private static final String JIRA_COMPONENTS_URL = JIRA_URL + '/rest/api/2/project/JENKINS/components'
+
+    abstract String getComponentId(String componentName);
+
+    /* Singleton support */
+    private static JiraAPI INSTANCE = null
+    static synchronized JiraAPI getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new JiraImpl()
+        }
+        return INSTANCE
+    }
+
+    private static class JiraImpl extends JiraAPI {
+
+        private Map<String, String> componentNamesToIds;
+
+        @SuppressFBWarnings("SE_NO_SERIALVERSIONID")
+        private void ensureDataLoaded() {
+            if (componentNamesToIds == null) {
+                componentNamesToIds = new HashMap<>();
+
+                LOGGER.log(Level.INFO, "Retrieving components from Jira...")
+                URL url = new URL(JIRA_COMPONENTS_URL)
+                HttpURLConnection conn = (HttpURLConnection) url.openConnection()
+
+                conn.setRequestMethod('GET')
+                conn.connect()
+                String text = conn.getInputStream().getText()
+
+                def json = new JsonSlurper().parseText(text)
+
+                json.each {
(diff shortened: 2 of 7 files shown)


halkeye:
I'm happy with this format. Its still explicit while still having less typing that tim wanted.

I looked over all the updators and everything and things looked good.

I would say the only thing I'd like to see after this is updating the CI to make sure it validates all new info coming in a PR, we've had issues where github urls or package names have repos or anything, but thats not something that should get done right now, just long term.

halkeye (approved):

oleg-nenashev (approved):
Looks good, thanks @daniel-beck !

daniel-beck:
> I would say the only thing I'd like to see after this is updating the CI to make sure it validates all new info coming in a PR

Agree and I considered it, but a lot of the possible validation depends on the data in an outside system and I don't want to introduce "random" failures by blocking e.g. when a Jira component isn't found (e.g. renamed), or when a GH issue tracker is disabled, or similar. That seems like a job for a new `-Werror` flag used only in CI.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/JiraAPI.groovy
```diff
@@ -0,0 +1,66 @@
...
+        private Map<String, String> componentNamesToIds;
+
+        @SuppressFBWarnings("SE_NO_SERIALVERSIONID")
+        private void ensureDataLoaded() {
+            if (componentNamesToIds == null) {
+                componentNamesToIds = new HashMap<>();
+
+                LOGGER.log(Level.INFO, "Retrieving components from Jira...")
+                URL url = new URL(JIRA_COMPONENTS_URL)
+                HttpURLConnection conn = (HttpURLConnection) url.openConnection()
```
timja, line 43: does this need closing? might not be such a big deal because it's short running

src/main/java/io/jenkins/infra/repository_permissions_updater/Definition.java
```diff
@@ -2,14 +2,138 @@
...
+                }
+            }
+            return jiraComponentId;
+        }
+
+        public String getViewUrl(JiraComponentSource source) {
+            if (isJira()) {
+                final String id = loadComponentId(source);
+                if (id != null) {
+                    return "https://issues.jenkins.io/issues/?jql=component=" + id;
```
timja, line 80: shouldn't this use the property for the jira url?

src/main/java/io/jenkins/infra/repository_permissions_updater/Definition.java
```diff
@@ -2,14 +2,138 @@
...
+        }
+
+        public String getReportUrl(JiraComponentSource source) {
+            if (!report) {
+                return null;
+            }
+            if (isJira()) {
+                final String id = loadComponentId(source);
+                if (id != null) {
+                    return "https://issues.jenkins.io/secure/CreateIssueDetails!init.jspa?pid=10172&issuetype=1&components=" + id;
```
timja, line 97: same here about property?

timja (approved):
no blockers looks great and solves all the issues I believe, nice work with the yaml references.

cc @slide we'll need to update the ircbot too

olblak (approved):

daniel-beck:
On-holding this to have an opportunity to amend it 😄

halkeye:
> On-holding this to have an opportunity to amend it 😄

If the amendments are tim's comments. I would vote merging and revisiting, none of them are blocking.

permissions/plugin-configuration-as-code.yml
```diff
@@ -12,3 +12,7 @@ developers:
 security:
   contacts:
     jira: "jetersen"
+issues:
+  - github: *GH
+  - jira: 23170 # configuration-as-code-plugin
```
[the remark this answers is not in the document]
daniel-beck: Yep. See readme for explanation of the tradeoffs.

(Internally this queries Jira to determine ID from name because ID is needed for 'New Issue' URLs, so either works functionality wise.)

timja:
@daniel-beck shall we 🚢  and improve later?

daniel-beck:
@timja I had an idea for an update to the output format, would rather like to get that in first. I hope to find time before the weekend. The regressions kept me busy recently 😢

daniel-beck:
Un-on-holding this. I've not found the time to do what I wanted to get in again and don't feel good blocking @halkeye further. I'll just have to adapt any downstream consumers in case of incompatible changes to the output format.

markyjackson-taulia (approved):

PR #1818, opened 2021-02-04
Allow incrementals publishing for Stapler

basil:
See [stapler/stapler#209 (comment)](https://github.com/stapler/stapler/pull/209#issuecomment-772621968). Stapler incrementals publishing regressed because `github.index.json` no longer has an entry for `stapler/stapler` (caused by #1806). I ensured that with this change `github.index.json` again contains an entry for `stapler/stapler`.

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/Definition.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/Definition.java
@@ -189,7 +189,7 @@ public class Definition {
     }
 
     public String getGithub() {
-        if (github != null && github.startsWith("jenkinsci/")) {
+        if (github != null && (github.startsWith("jenkinsci/") || github.startsWith("stapler/"))) {
             return github;
         }
         return null;


markyjackson-taulia (approved):

timja (approved):

jglick (commented):
There is also `isGitHubIssues` hard-coding `jenkinsci`.

I do not follow why the check for the org prefix was introduced in #1806 at all.

jglick:
@daniel-beck ^^^?

timja:
I think it was along the lines of not showing issue trackers that we don't have some control over.

i.e. if objectionable content shows up, and we scrape it and show it on the website. We have control over our own Jira and GitHub org but not other ones.

Like we don't accept plugin docs hosted on GitHub outside of our org.

The `isGitHubIssues` shouldn't affect stapler as it's mostly for Jenkins core and the plugin site to know where the issue tracker is, I can't see stapler needing that

jglick:
Note that @oleg-nenashev had some plan, long paused, to move Stapler and other components into @jenkinsci, which would make these sorts of things simpler.

daniel-beck:
> I do not follow why the check for the org prefix was introduced in #1806 at all.

> I think it was along the lines of not showing issue trackers that we don't have some control over.

Jenkins plugins are expected to be hosted in the `jenkinsci` org. We're not actively removing plugins that got in before this was actively enforced through the hosting process, but we're not supporting outside hosting either.

PR #1856, opened 2021-03-12
[INFRA-2835] Ensure that users exist in Jira

daniel-beck:
[INFRA-2835](https://issues.jenkins.io/browse/INFRA-2835)

- Use a JSON report file as a cache
- For any user not specified in there, query Jira directly

----

Downstream from https://github.com/jenkins-infra/infra-reports/pull/26

I'm not really happy with this solution, longer term we probably want to create accounts (both Jira and Artifactory) for users lacking one automatically. Meanwhile this seems like the best approach as an extension of the current behavior.

Needs new credential `jiraUser` of any Jira account (no admin permissions) both on ci.j.io and trusted.ci (already added).

To make this work, I created a dozen or so accounts in Jira for users who didn't (see https://github.com/jenkins-infra/repository-permissions-updater/pull/1856/commits/a3de6a7682cde49ca0d4e38de72eea6b527c5302 for affected accounts)

### DIFF
--- a/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
+++ b/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
@@ -150,7 +150,7 @@ class ArtifactoryPermissionsUpdater {
                             def existsInJira = KnownUsers.existsInJira(developer) || JiraAPI.getInstance().isUserPresent(developer)
 
                             if (!existsInArtifactory && !existsInJira) {
-                                reportChecksApiDetails(developer + " needs to log in to Artifactory",
+                                reportChecksApiDetails(developer + " needs to log in to Artifactory and Jira",
                                 """
                                 ${developer} needs to log in to [Artifactory](https://repo.jenkins-ci.org/) and [Jira](https://issues.jenkins.io/).
 
@@ -176,7 +176,7 @@ class ArtifactoryPermissionsUpdater {
                             }
 
                             if (!existsInJira) {
-                                reportChecksApiDetails(developer + " needs to log in to Artifactory",
+                                reportChecksApiDetails(developer + " needs to log in to Jira",
                                         """
                                 ${developer} needs to log in to [Jira](https://issues.jenkins.io/).
                                 """.stripIndent())
--- a/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
+++ b/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
@@ -146,17 +146,41 @@ class ArtifactoryPermissionsUpdater {
                         users [:]
                     } else {
                         users definition.developers.collectEntries { developer ->
-                            if (!KnownUsers.exists(developer.toLowerCase())) {
-                                reportChecksApiDetails(developer + " needs to login to artifactory", 
-                                """
-                                ${developer} needs to login to [artifactory](https://repo.jenkins-ci.org/).
+                            def existsInArtifactory = KnownUsers.existsInArtifactory(developer)
+                            def existsInJira = KnownUsers.existsInJira(developer) || JiraAPI.getInstance().isUserPresent(developer)
 
-                                We resync our artifactory list hourly, so you will need to wait some time before rebuilding your pull request.
+                            if (!existsInArtifactory && !existsInJira) {
+                                reportChecksApiDetails(developer + " needs to log in to Artifactory",
+                                """
+                                ${developer} needs to log in to [Artifactory](https://repo.jenkins-ci.org/) and [Jira](https://issues.jenkins.io/).
+
+                                We resync our Artifactory list hourly, so you will need to wait some time before rebuilding your pull request.
+                                The easiest way to trigger a rebuild is to close your pull request, wait a few seconds and then reopen it.
+
+                                Alternatively the hosting team can re-trigger it if you post a comment saying you have now logged in.
+                                """.stripIndent())
+                                throw new IllegalStateException("User name not known to Artifactory and Jira: " + developer)
+                            }
+
+                            if (!existsInArtifactory) {
+                                reportChecksApiDetails(developer + " needs to log in to Artifactory",
+                                        """
+                                ${developer} needs to log in to [Artifactory](https://repo.jenkins-ci.org/).
+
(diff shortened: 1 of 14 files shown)


src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -146,18 +146,42 @@ class ArtifactoryPermissionsUpdater {
...
+                                We resync our Artifactory list hourly, so you will need to wait some time before rebuilding your pull request.
                                 The easiest way to trigger a rebuild is to close your pull request, wait a few seconds and then reopen it.
 
                                 Alternatively the hosting team can re-trigger it if you post a comment saying you have now logged in.
                                 """.stripIndent())
                                 throw new IllegalStateException("User name not known to Artifactory: " + developer)
                             }
+
+                            if (!existsInJira) {
+                                reportChecksApiDetails(developer + " needs to log in to Artifactory",
```
timja, line 179:
```suggestion
                                reportChecksApiDetails(developer + " needs to log in to Jira",
```

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -146,18 +146,42 @@ class ArtifactoryPermissionsUpdater {
...
+                                        """
+                                ${developer} needs to log in to [Jira](https://issues.jenkins.io/).
```
timja, line 181: shouldn't this have the same / similar instructions as artifactory, saying how often we sync the list and how to rebuild the pull request?

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -146,18 +146,42 @@ class ArtifactoryPermissionsUpdater {
...
+
+                                We resync our Artifactory list hourly, so you will need to wait some time before rebuilding your pull request.
                                 The easiest way to trigger a rebuild is to close your pull request, wait a few seconds and then reopen it.
 
                                 Alternatively the hosting team can re-trigger it if you post a comment saying you have now logged in.
                                 """.stripIndent())
                                 throw new IllegalStateException("User name not known to Artifactory: " + developer)
                             }
+
+                            if (!existsInJira) {
```
timja, line 178: shouldn't we re-work this to check both jira and artifactory before throwing an exception and also produce a  different checks report if both are missing?

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
daniel-beck: No, because we only use that list as a cache. Regular user credentials are enough to query live, so that's what we do here.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
daniel-beck: https://github.com/jenkins-infra/repository-permissions-updater/pull/1856/files#diff-3e4cb1ddb9f0646c5ddc1302282ba93e58eb6d012fab254ecd90cad8b7202283R155

timja: ah i missed the first check

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -146,18 +146,42 @@ class ArtifactoryPermissionsUpdater {
                         users [:]
                     } else {
                         users definition.developers.collectEntries { developer ->
-                            if (!KnownUsers.exists(developer.toLowerCase())) {
-                                reportChecksApiDetails(developer + " needs to login to artifactory", 
+                            def existsInArtifactory = KnownUsers.existsInArtifactory(developer)
+                            def existsInJira = KnownUsers.existsInJira(developer) || JiraAPI.getInstance().isUserPresent(developer)
+
+                            if (!existsInArtifactory && !existsInJira) {
+                                reportChecksApiDetails(developer + " needs to log in to Artifactory",
```
timja, line 153: hopefully & is a valid character, worried about the length as it will get cut off if it's too long, thoughts? easy to test after merge
```suggestion
                                reportChecksApiDetails(developer + " needs to log in to Artifactory & Jira",
```

daniel-beck: Why not sidestep the problem and go with

```suggestion
                                reportChecksApiDetails(developer + " needs to log in to Artifactory and Jira",
```

timja: sure go for it, let's see what it looks like in the UI after merge

daniel-beck: Will amend when I update the report URL.

timja (approved):

daniel-beck:
This is now ready and could be merged as is.

Tempted to just create accounts in Jira for affected users and revert the permissions file changes. Thoughts?

timja:
> This is now ready and could be merged as is.
> 
> Tempted to just create accounts in Jira for affected users and revert the permissions file changes. Thoughts?

If it's easy to do so and you have the time then yes that would be the preferred approach

oleg-nenashev (approved):
Looks good 👍

daniel-beck:
> > This is now ready and could be merged as is.
> > Tempted to just create accounts in Jira for affected users and revert the permissions file changes. Thoughts?
> 
> If it's easy to do so and you have the time then yes that would be the preferred approach

Done. I backed out all permissions file changes.

timja (approved):

daniel-beck:
CC @slide bot created PR messages in this repo should probably be updated accordingly.

daniel-beck:
Thanks @timja @oleg-nenashev for your timely reviews ❤️

PR #1880, opened 2021-03-14
[WEBSITE-658] Add maintainers index

daniel-beck:
See [WEBSITE-658](https://issues.jenkins.io/browse/WEBSITE-658)

By consuming this file, update-center2 will be able to provide up-to-date maintainer information based on actual upload permissions, rather than the usually very badly maintained information from the `pom.xml` file.

Selected sample output:

```
{
    "org.jenkins-ci.plugins:matrix-auth": [
        "danielbeck"
    ],
    "org.jenkins-ci.plugins:strict-crumb-issuer": [
        "danielbeck",
        "wfollonier"
    ]
}
```

(Accidental origin branch, sorry about that. Please delete on merge.)

### DIFF
--- a/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
+++ b/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
@@ -59,6 +59,7 @@ class ArtifactoryPermissionsUpdater {
         Map<String, Set<String>> pathsByGithub = new TreeMap()
         Map<String, List> issueTrackersByPlugin = new TreeMap()
         Map<String, List<Definition>> cdEnabledComponentsByGitHub = new TreeMap<>()
+        Map<String, List<String>> maintainersByComponent = new HashMap<>()
 
         yamlSourceDirectory.eachFile { file ->
             if (!file.name.endsWith('.yml')) {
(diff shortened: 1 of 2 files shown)


markyjackson-taulia (approved):

MarkEWaite:
Is there any harm in a plugin having a large list of maintainers, mostly inactive?  

Should I be removing upload permissions from inactive maintainers of plugins that I maintain?

daniel-beck:
> Is there any harm in a plugin having a large list of maintainers, mostly inactive?

Chaos during development when an unexpected release shows up, accidental releases, and potentially malicious releases by taking over dormant maintainer accounts, e.g. using password spraying. More accounts are more chances of problems.

> Should I be removing upload permissions from inactive maintainers of plugins that I maintain?

IMO, strong yes. See #782 for when I did that with matrix-auth not long after I took over.

(For the "maintainers list" on the plugin site, an argument can be made that this is also highlighting past contributors, rather than just "these are the folks who currently can create releases". The former argument is probably better served by integrating with GitHub Pulse or a similar system showing contributions IMO.)

My best guess why this isn't happening regularly is that it's not particularly well documented and/or that folks don't want to step on toes of past contributors who may have a less utilitarian view of being present on the maintainers list.

uhafner:
That means a lot of cleanup needs to be done. I am also in the permissions file of several plugins that I released some time in the past. Maybe it makes sense to highlight the desired approach in the mailing list so everybody can try to minimize his responsibilities.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -121,6 +122,25 @@ class ArtifactoryPermissionsUpdater {
...
+
+                String key = groupId + ":" + artifactId
+
+                if (maintainersByComponent.containsKey(key)) {
+                    LOGGER.log(Level.WARNING, "Duplicate maintainers entry for component: " + key)
+                }
+                // A potential issue with this implementation is that groupId changes will result in lack of maintainer information for the old groupId.
+                // In practice this will probably not be a problem when path changes here and subsequent release are close enough in time.
+                // Alternatively, always keep the old groupId around for a while.
+                maintainersByComponent.computeIfAbsent(key, { _ -> new ArrayList<>(Arrays.asList(definition.developers)) })
```
timja, line 141: thoughts on pulling out display names for plugin site? we're changing from:

> Tim Jacomb

to:
> timja

with this change

timja (approved):

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
daniel-beck: If we apply this directly it would be worse, yes. My current plan is to add a new https://github.com/jenkins-infra/infra-reports/ that provides a mapping from user ID to user name (perhaps from Jira, to not require LDAP access) for maintainer accounts. TBD whether that'll make it into the first iteration, or will be a future enhancement. [INFRA-2926](https://issues.jenkins.io/browse/INFRA-2926) is a bit annoying in this context though.

oleg-nenashev:
FTR this addresses https://issues.jenkins.io/browse/WEBSITE-658 (wrong project?)

oleg-nenashev (approved):
Looks good! Thanks for handling it @daniel-beck !

daniel-beck:
@oleg-nenashev Thanks for the reference, I forgot that existed.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
daniel-beck: @timja @oleg-nenashev https://github.com/jenkins-infra/infra-reports/pull/27

PR #1938, opened 2021-04-27
Use ant matcher for artifactory path in bom

timja:
https://github.com/jenkins-infra/repository-permissions-updater/pull/1937#issuecomment-827440676

@jetersen @daniel-beck @jglick


daniel-beck:
Oh, that looks nice. If I don't have validation against this, it should work. I recommend reviewing the generated payload for any transformation issues, but otherwise 👍

timja:
> Oh, that looks nice. If I don't have validation against this, it should work. I recommend reviewing the generated payload for any transformation issues, but otherwise 👍

let's give it a go, can easily revert it

daniel-beck:
This changes `github.index.json`. Do we use that for incrementals for the bom?

timja:
> This changes `github.index.json`. Do we use that for incrementals for the bom?

Yes, may need adapting there, don't know that anyone actually uses incremental bom builds though, @jglick?

jglick:
Cannot recall offhand if https://github.com/jenkins-infra/incrementals-publisher/blob/00be8881121de3bb26ba91605aafedeecbba7435/lib/permissions.js relies on this field. I would recommend reverting this.

timja:
why?

jglick:
Could break incremental deployment at least of this component, which we want to enable.

timja:
I'll fix forward if required, I was able to do a release with this config ^^,

https://github.com/jenkinsci/bom/pull/498

jetersen:
@timja https://github.com/jenkins-infra/incrementals-publisher/blob/00be8881121de3bb26ba91605aafedeecbba7435/lib/permissions.js#L19 this line would require fixing, if I am reading the code correctly 😅

jglick:
Rather https://github.com/jenkins-infra/incrementals-publisher/blob/00be8881121de3bb26ba91605aafedeecbba7435/lib/permissions.js#L30 looks like it would break against

```json
    "jenkinsci/bom": [
        "io/jenkins/tools/bom/bom-*"
    ],
```

jetersen:
@jglick right I simply cannot read 🙈

timja:
Well bom builds themselves keep flaking out but I did a PR anyway as a unit test showed it wouldn't work:
https://github.com/jenkins-infra/incrementals-publisher/pull/16

jetersen:
@timja so this https://github.com/jenkins-infra/incrementals-publisher/releases/tag/v1.4.0 fixes the incrementals publisher on Jenkins CI right?

timja:
Yes it did and 👍

PR #2073, opened 2021-09-04
Enable cd for nomad-plugin

multani:
I'd like to enabled automated releases for the [Nomad plugin](https://github.com/jenkinsci/nomad-plugin/).

Following: https://www.jenkins.io/doc/developer/publishing/releasing-cd/

Thanks!


timja:
releases are disabled atm so this won't work for a bit: 
https://www.jenkins.io/blog/2021/09/04/wiki-attacked/

PR #2109, opened 2021-10-03
Fix path for `jenkins-maven-plugin`

basil:
I just tried to do a release of Maven Repository and got this error:

```
[INFO] --- maven-deploy-plugin:2.8.2:deploy (default-deploy) @ jenkins-maven-plugin ---
Uploading to maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/1.5/jenkins-maven-plugin-1.5.jar
Uploaded to maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/1.5/jenkins-maven-plugin-1.5.jar (20 kB at 13 kB/s)
Uploading to maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/1.5/jenkins-maven-plugin-1.5.pom
Uploaded to maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/1.5/jenkins-maven-plugin-1.5.pom (3.5 kB at 5.2 kB/s)
Downloading from maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/maven-metadata.xml
Downloaded from maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/maven-metadata.xml (660 B at 3.1 kB/s)
Downloading from maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/maven-metadata.xml
Downloaded from maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/maven-metadata.xml (248 B at 779 B/s)
Uploading to maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/maven-metadata.xml
Uploaded to maven.jenkins-ci.org: https://repo.jenkins-ci.org/releases/com/nirima/jenkins/repository/jenkins-maven-plugin/maven-metadata.xml (614 B at 990 B/s)
[...]

PR #2111, opened 2021-10-03
Add permissions for Maven Repository plugin parent POM

basil:
Trying to fix [this build failure](https://basilcrow.com/log2.txt). Looking at other examples, it seems that adding `com/nirima/jenkins/repository/pom` to `component-jenkins-maven-plugin.yml` in #2109 wasn't quite correct. Other examples have a dedicated `pom-${PLUGIN}.xml` file that (crucially) contains the `artifactId` of the POM as its `name`. So in this PR I am adding a `pom-maven-repository-plugin.yml` whose name is `pom`, which matches the `artifactId` from [here](https://github.com/jenkinsci/maven-repository-plugin/blob/3bfad5f403e45ed9d9e48e107657f97a8c993d70/pom.xml#L12). So I think this should resolve my error.

While I was here, I added GitHub URLs for all three of this plugin's YAML files, and I updated the main plugin YAML file with proper issue tracking.

CC @timja

### DIFF
--- a/permissions/component-jenkins-maven-plugin.yml
+++ b/permissions/component-jenkins-maven-plugin.yml
@@ -1,9 +1,9 @@
 ---
 name: "jenkins-maven-plugin"
+github: "jenkinsci/maven-repository-plugin"
 paths:
 - "com/nirima/jenkins/repository"
 - "com/nirima/jenkins/repository/jenkins-maven-plugin"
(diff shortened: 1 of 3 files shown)


timja (approved):

basil:
It worked! Thanks, @timja!

PR #2154, opened 2021-11-01
[INFRA-3113] Add a team concept

batmat:
(PR on the tooling, **not** on the permissions per-se. See [INFRA-3113](https://issues.jenkins-ci.org/browse/INFRA-3113))

Now one can
1) define a team inside the `teams/` directory
2) reference this team from a file under `permissions/`.
   To do so, one uses `@` prefix with the team name defined in the yml.

This feature addition is accompanied by a handful of half-random changes in the permissions files to demonstrate the feature.

### locally, if need be:

```sh
wget https://reports.jenkins.io/artifactory-ldap-users-report.json
wget https://reports.jenkins.io/jira-users-report.json
java -jar -DdefinitionsDir=$PWD/permissions \
          -DartifactoryApiTempDir=$PWD/json \
          -DartifactoryUserNamesJsonListUrl=file://$PWD/artifactory-ldap-users-report.json \
          -DjiraUserNamesJsonListUrl=file://$PWD/jira-users-report.json \
          -Djava.util.logging.SimpleFormatter.format='%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS %4$s: %5$s%6$s%n' \
          -DdryRun=true \
          -jar target/repository-permissions-updater-*-bin/repository-permissions-updater-*.jar
```

### From the CI
Download archived artifacts, check the `json/maintainers.index.json` content.
E.g.

```sh
cat json/maintainers.index.json | jq -r .'["org.jenkins-ci.plugins:credentials"]'
[
  "jglick",
  "kohsuke",
  "oleg_nenashev",
  "jvz",
  "jthompson",
  "danielbeck",
  "wfollonier"
]
```

(this content is expected to be the same with or without this PR, potentially just with a different order).

cc @jtnord


teams/team-cloudbees.yml
```diff
@@ -0,0 +1,11 @@
+---
+name: "cloudbees-developers"
```
jtnord, line 2: Would it not make sense to use the same name as the file? (Also should the file start with "team-" as it is in the teams folder and team is obvious from the `@name` notation?

batmat: Yeah. I don't disagree actually. The filename should `xyz.yml` as it is currently.
I thought about dropping the `name` attribute from the team yaml, and using the filename as the team name too. 
The reason why I didn't go that path (yet) is that I was unsure we'd not want a team name with characters that wouldn't fit well a filename.

Not fully decided yet. I'll _literally_ sleep on it :-).

What about the rest?

timja: better to have the name in the yaml file imo, more obvious and flexible I think.

I agree no need to have team in the file name, but not a blocker

timja (approved):

teams/team-cloudbees.yml
jtnord: > better to have the name in the yaml file imo, more obvious

I would say it is more obvious when you have the file open - but less obvious if you want to have a look for and edit a file for a given team (as there is nothing ensuring they match).

hopefully there won't be too many of these files(?) so having a different team name from the file name will not be an issue.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -42,6 +42,60 @@ class ArtifactoryPermissionsUpdater {
...
+    private static void expandTeams(Definition definition, Map<String, Set<TeamDefinition>> teamsByName) {
+        Set<String> developers = definition.developers
+        Set<String> expandedDevelopers = [] as Set
+
+        developers.each { developerName ->
+            if (developerName.startsWith('@')) {
+                String teamName = developerName.substring(1)
+                Set teamDevs = teamsByName.get(teamName)?.developers
+                if (teamDevs == null ) {
+                    throw new Exception("Team $developerName not found!")
```
jtnord, line 85:
```suggestion
                    throw new Exception("Team $teamName not found!")
```

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -42,6 +42,60 @@ class ArtifactoryPermissionsUpdater {
...
+                }
+                if (teamDevs.isEmpty()) {
+                    throw new Exception("Team $developerName is empty?!")
```
jtnord, line 88:
```suggestion
                    throw new Exception("Team $teamName is empty?!")
```

teams/team-cloudbees.yml
batmat: I suppose I can add a check to enforce the `name` of the filename do match, at least for now. I don't like the repetition, but well I can live with it.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -42,6 +43,66 @@ class ArtifactoryPermissionsUpdater {
...
+        File teamsDir = new File('teams/')
+
+        Set<TeamDefinition> teams = [] as Set
+
+        teamsDir.eachFile { teamFile ->
+            try {
+                TeamDefinition newTeam = yaml.loadAs(new FileReader(teamFile), TeamDefinition.class)
+                teams.add(newTeam)
+
+                String expectedName = "team-${newTeam.name}.yml"
```
timja, line 61: makes more sense to me since you're in the team folder
```suggestion
                String expectedName = "${newTeam.name}.yml"
```

batmat: yeah, I thought that we may want to add different files in that folder. But for now we can simplify, let me push a fix for this just now.

batmat: done

batmat:
I've marked this PR ready for review (i.e. not draft anymore).

AFAICT, given Tim's approval, I plan to merge soon later today.

Feel free to provide feedback, even after I have merged. I'll be happy to consider adjusting as need be.

teams/team-cloudbees.yml
batmat: done

teams/security.yml
```diff
@@ -0,0 +1,5 @@
+---
+name: "security" # using a different name than the file just to show this is the value that is used in the other files
```
jtnord, line 2:
```suggestion
name: "security"
```

jtnord (approved):
Code kooks ok to me.  One comment in the YAML is no longer correct so could do with an update before merging to reduce confusion but otherwise LGTM!

batmat:
@daniel-beck do you still plan to review this? Danke

batmat:
Moving on as written previously anyways we can easily adjust after the fact of there's feedback.

daniel-beck:
@batmat 

> do you still plan to review this?

Yes, as I wrote in https://cloudbees.slack.com/archives/CJEUVT67P/p1636025133071200?thread_ts=1636025081.071100&cid=CJEUVT67P

batmat:
Feel free to provide feedback. I'll address it in a follow-up as need be. Thank you

batmat:
And re-reading your feedback from last week, I realize that yes I could've waited a bit more sorry  For some reason my mind remembered the first part only in 'later this week _or next week_'.

PR #2209, opened 2021-11-28
Issues link should include open issues only

zbynek:
# Description

Fixes https://github.com/jenkins-infra/plugin-site/issues/853 using @timja 's suggestion for JQL change.

CC @daniel-beck @MarkEWaite

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/Definition.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/Definition.java
@@ -77,7 +77,7 @@ public class Definition {
             if (isJira()) {
                 final String id = loadComponentId(source);
                 if (id != null) {
-                    return "https://issues.jenkins.io/issues/?jql=component=" + id;
+                    return "https://issues.jenkins.io/issues/?jql=resolution is EMPTY and component=" + id;
                 }
                 return null;
             }


MarkEWaite (approved):
I like the change very much though I am a little concerned that `getViewUrl` might be misinterpreted as a method name to refer to the URL that will show all issues.  Should the javadoc be updated to note that the returned URL shows open issues rather than all issues?

The GitHub issues link already shows only open issues.  I think this makes the Jira and GitHub links have similar results.

timja (approved):
Assuming it doesn’t need to be url encoded

MarkEWaite:
> Assuming it doesn’t need to be url encoded

When I open `https://issues.jenkins.io/issues/?jql=resolution is EMPTY and component=15543`, the expected result appears and Google Chrome rewrites the URL as `https://issues.jenkins.io/issues/?jql=resolution%20is%20EMPTY%20and%20component%3D15543`.  I'm not sure if that indicates that the string needs to be URL encoded or not...

zbynek:
I think URL encoding should be up to the consumer, if needed at all.

daniel-beck:
[RFC 1738](https://datatracker.ietf.org/doc/html/rfc1738#section-2.2):

> Octets must be encoded if they have no corresponding graphic character within the US-ASCII coded character set, if the use of the corresponding character is unsafe, or if the corresponding character is reserved for some other interpretation within the particular URL scheme.
> …
> Characters can be unsafe for a number of reasons.  The space character is unsafe because significant spaces may disappear and insignificant spaces may be introduced when URLs are transcribed or typeset or subjected to the treatment of word-processing programs.

halkeye:
It's hard to encode parameters after the fact. You'll have to split on ? And = and in this case since there is already = in the string it'll be rough.

You should encode the jql

daniel-beck:
If we go with this change (which per https://github.com/jenkins-infra/plugin-site/issues/853#issuecomment-981126881 I'm not sure about), docs need to be updated as well to specify this only links to open issues.

zbynek:
Closing in favor of front-end solution.

PR #2284, opened 2022-01-01
Add hosting verifier (ported from ircbot)

timja:
Ports https://github.com/jenkins-infra/ircbot/tree/main/src/main/java/org/jenkinsci/backend/ircbot/hosting to GitHub issues.

Minimal changes done just:
- Refactor Jira classes out to a plain object for storing field information
- Change Jira markup to GitHub markdown
- Introduced a Jenkins project checker which forces all users to log into Jira and artifactory before hosting is done
- Switched to gmavenplus, groovy and java interop wasn't working for me and this fixed it
- Commented out 'updater' functionality i.e. removing `-jenkins`, may look at it later but it complicates things a bit

There's plenty of other commented out code, it was already like this, no plans to address right now

A follow up PR will add a GitHub workflow that runs this on Issue open and via comment ops

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci</groupId>
         <artifactId>jenkins</artifactId>
-        <version>1.55</version>
+        <version>1.69</version>
     </parent>
 
     <groupId>io.jenkins.infra</groupId>
@@ -27,13 +27,22 @@
                 </configuration>
             </plugin>
             <plugin><!-- run unit test in src/test/java -->
-                <groupId>org.codehaus.gmaven</groupId>
-                <artifactId>gmaven-plugin</artifactId>
+                <groupId>org.codehaus.gmavenplus</groupId>
+                <artifactId>gmavenplus-plugin</artifactId>
+                <version>1.13.1</version>
                 <!-- version specified in grandparent pom -->
                 <executions>
                     <execution>
                         <goals>
+                            <goal>execute</goal>
+                            <goal>addSources</goal>
+                            <goal>addTestSources</goal>
+                            <goal>generateStubs</goal>
                             <goal>compile</goal>
+                            <goal>generateTestStubs</goal>
+                            <goal>compileTests</goal>
+                            <goal>removeStubs</goal>
+                            <goal>removeTestStubs</goal>
                         </goals>
                     </execution>
                 </executions>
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -27,21 +27,32 @@
                 </configuration>
             </plugin>
             <plugin><!-- run unit test in src/test/java -->
-                <groupId>org.codehaus.gmaven</groupId>
-                <artifactId>gmaven-plugin</artifactId>
+                <groupId>org.codehaus.gmavenplus</groupId>
```
halkeye, line 30: i have to look into this eventually, but what does gmaven plus do?

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/JenkinsProjectUserVerifier.java
```diff
@@ -0,0 +1,29 @@
...
+
+        String missingInArtifactory = request.getJenkinsProjectUsers()
+                .stream().filter(user -> !KnownUsers.existsInArtifactory(user))
+                .collect(Collectors.joining(", "));
+        String missingInJira = request.getJenkinsProjectUsers()
+                .stream().filter(user -> !KnownUsers.existsInJira(user))
+                .collect(Collectors.joining(", "));
+
+       if (StringUtils.isNotBlank(missingInArtifactory)) {
+           hostingIssues.add(new VerificationMessage(VerificationMessage.Severity.REQUIRED, "The following usernames in 'Jenkins project users to have release permission' need to log into [Artifactory](https://repo.jenkins-ci.org/): %s (reports are re-synced hourly, wait to re-check for a bit after logging in)", missingInArtifactory));
```
halkeye, line 22: might be worth mentioning they can self check via https://reports.jenkins.io/artifactory-ldap-users-report.json ?

halkeye:
seems fine to me. Do youw want to add the package + name length checker  thing too? or leave it for future?
We can run this and jira in parallel right?

halkeye (approved):

pom.xml
timja: It’s a groovy compiler, successor to gmaven, Java / groovy interop wasn’t working fully without using the more modern plus rewrite. 

I wanted to reuse the code that checks a user has logged into Jira and Artifactory

timja:
> seems fine to me. Do youw want to add the package + name length checker thing too? or leave it for future?

I did add it :) it’s in the maven verifier, didn’t add to Gradle but meh barely get any of those

> We can run this and jira in parallel right?

yeah we can , I haven’t touched any of the ircbot code

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/JenkinsProjectUserVerifier.java
timja: Could do but let’s see how it goes the messages are already quite long, may need some iteration

PR #2326, opened 2022-01-20
Chore(Jenkinsfile) use a maven-11 agent and stop relying on pipeline tools

dduportal:
Using tools is blocking a lot of PRs. The proposal is to stop relying on tools when both maven and java are already present in the machines.

Tested with a "debug" commit that shall be removed in https://github.com/jenkins-infra/repository-permissions-updater/pull/2315 with the help of @tfennelly .

Is there any impact of doing this that I could have forgotten?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ props += pipelineTriggers(triggers)
 properties(props)
 
 
-node('java') {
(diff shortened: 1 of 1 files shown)


timja (approved):
thanks

PR #2327, opened 2022-01-20
Add java label, trusted doesn't have maven-* labels

timja:
Relates to https://github.com/jenkins-infra/helpdesk/issues/2754

Trusted CI doesn't have maven-* or java-* labels only java

According to https://github.com/jenkins-infra/packer-images/search?q=DEFAULT_JDK

Default JDK is 11

Which should be equivalent to this

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ props += pipelineTriggers(triggers)
 properties(props)
 
 
-node('maven-11') {
+node('java') {
     try {
         stage ('Clean') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -27,7 +27,7 @@ props += pipelineTriggers(triggers)
 properties(props)
 
 
-node('maven-11') {
+node('java') {
```
timja, line 30:
```suggestion
node('maven-11 || java') {
```

PR #2449, opened 2022-03-18
Migrate strict-crumb-issuer to 'security' team

daniel-beck:
This effectively adds @Kevin-CB as a new co-maintainer.

@Wadeck to approve.

A potential problem here is the semantic conflict of `security` being either "folks working pretty much full time in the Jenkins security team led by the security officer" or "the people in the team at CloudBees working on Jenkins security", which overlap but are not identical. I'll leave that for @Wadeck to clear up, potentially with @batmat who defined the team.


Wadeck (approved):
👍 
No need to define more precisely at this point.

Issue #2316, opened 2022-01-13
Hosting checker didn't add all people with release permissions


timja:
### Reproduction steps

1. https://github.com/jenkins-infra/repository-permissions-updater/issues/2314
2. https://github.com/jenkins-infra/repository-permissions-updater/pull/2315
3. https://github.com/jenkins-infra/repository-permissions-updater/pull/2472

### Expected Results

tfennelly
agentgonzo

### Actual Results

tfennelly

Issue #2550, opened 2022-05-17
Worksoft Continuous Testing Manager Plugin


raleigh-rinehart:
### Repository URL

https://github.com/Worksoft/worksoft-ctm-plugin

### New Repository Name

ws-ctm-plugin

### Description

This is the official Worksoft Supported Plugin for CTM.
The plugin adds build steps for interacting with a Worksoft's Continuous Testing Manager to run your Certify processes or ALM labs.

### GitHub users to have commit permission

@raleigh-rinehart
@glenpgillman

### Jenkins project users to have release permission

rrinehart
glenpgillman

### Issue tracker

Jira

glenpgillman:
why is the hosting checker trying to validate a destination git repo that doesn't exist yet?   I'm confused.   Of course 'ws-ctm-plugin' will not have been created as a destination jenkins CI repo until this issue is approved.  Are we missing something ?

raleigh-rinehart:
At this point we are unsure what the problem is nor how to fix it to proceed.

The bot seems to be complaining about the destination repo `ws-ctm-plugin` not being valid.  Which of course it is not as it has not been created yet.  A search in Jenkins CI GitHub org for other repositories with the same name turns up empty, so that is not the issue either. 
Review the bot's actions (https://github.com/jenkins-infra/repository-permissions-updater/runs/6477610395?check_suite_focus=true) turns up nothing that points to the problem

@glenpgillman FYI

raleigh-rinehart:
@timja It looks like there might be a bug in `GradleVerifier.java`
Lines 70 & 71 seem to have the `from` repo and `to` repo swapped:
```java
   forkTo = issue.getRepositoryUrl();
   forkFrom = issue.getNewRepoName();
```

timja:
Ha thanks, makes sense, this may be the first gradle based plugin since this was introduced

timja:
Hello from your friendly Jenkins Hosting Checker

It appears you have some issues with your hosting request. Please see the list below and correct all issues marked Required. Your hosting request will not be approved until these issues are corrected. Issues marked with Warning or Info are just recommendations and will not stall the hosting process.
* :no_entry: Required: Repository URL 'ws-ctm-plugin' is not a valid GitHub repository (check that you do not have .git at the end, GitHub API doesn't support this).

You can re-trigger a check by editing your hosting request or by commenting `/hosting re-check`

timja:
[...]

PR #2551, opened 2022-05-17
Fix gradle verifier and add all people to release permission

timja:
Fixes https://github.com/jenkins-infra/repository-permissions-updater/issues/2316

see https://github.com/jenkins-infra/repository-permissions-updater/issues/2550#issuecomment-1129285522


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/GradleVerifier.java
```diff
@@ -67,11 +67,11 @@ public class GradleVerifier extends CodeVisitorSupport implements BuildSystemVer
     @Override
     public void verify(HostingRequest issue, HashSet<VerificationMessage> hostingIssues) throws IOException {
         GitHub github = GitHub.connect();
-        forkTo = issue.getRepositoryUrl();
-        forkFrom = issue.getNewRepoName();
+        forkFrom = issue.getRepositoryUrl();
```
halkeye, line 70: Haha oops

It was interesting to read as we return that exact same error in 3 different but similar flows

halkeye (approved):

PR #2596, opened 2022-06-15
Add hosting to assignees of new plugin requests

halkeye:
that way @jenkins-infra/hosting actually gets notified of new issues.

I suspect this won't work because you can't assign to a team.


lemeurherve:
> I suspect this won't work because you can't assign to a team.

Yes it shouldn't work. But an auto-notifier can be added like the one I finally got working on helpdesk: https://github.com/jenkins-infra/helpdesk/blob/main/.github/workflows/autonotifier.yaml

WDYT?

lemeurherve:
Another alternative, this GitHub app allows assigning a number of team members to issues: https://github.com/marketplace/actions/auto-assign-issue

PR #2678, opened 2022-07-27
Document JEP-229 CD and what this means for security

daniel-beck:
Needs @Wadeck approval.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,27 @@ Rename and edit the existing permissions file, changing both `name` and the last
 
 Change the `paths` to match the new Maven coordinates, or, if further uploads for the old coordinates are expected, add a new list entry.
 
+Managing Continuous Delivery (JEP-229 CD)
+-----------------------------------------
+
+Jenkins plugins and other components can be continuously delivered through a supported process described in [JEP-229](https://github.com/jenkinsci/jep/blob/master/jep/229/README.adoc).
+
+You can enable JEP-229 CD for your component by adding the following to your component's YAML file:
+
+```yaml
+cd:
+  enabled: true
+```
+
+**NOTE:** As of July 2022, this does not work for multimodule projects deploying multiple components, as these would have multiple YAML files in this repository.
+
+**IMPORTANT:**
+When using JEP-229 CD, every committer to your repository can create new releases by merging pull requests.
+As a result, the list of maintainer accounts maintained in your plugin's YAML file is no longer the single reference on who can publish new releases.
+The users listed there still serve as the contacts for security issues and plugin/component governance questions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,6 +75,27 @@ Rename and edit the existing permissions file, changing both `name` and the last
...
+  enabled: true
+```
+
+**NOTE:** As of July 2022, this does not work for multimodule projects deploying multiple components, as these would have multiple YAML files in this repository.
+
+**IMPORTANT:**
+When using JEP-229 CD, every committer to your repository can create new releases by merging pull requests.
+As a result, the list of maintainer accounts maintained in your plugin's YAML file is no longer the single reference on who can publish new releases.
+The users listed there still serve as the contacts for security issues and plugin/component governance questions.
+In particular, the Jenkins security team will _not_ make an effort to reach out to GitHub committers when maintainers are unresponsive before [announcing vulnerabilities without a fix](https://www.jenkins.io/security/plugins/#unresolved).
```
daniel-beck, line 96:
```suggestion
In particular, the Jenkins security team will _not_ make an effort to reach out to GitHub committers when maintainers (and security contacts, see below) are unresponsive before [announcing vulnerabilities without a fix](https://www.jenkins.io/security/plugins/#unresolved).
```

README.md
```diff
@@ -75,6 +75,27 @@ Rename and edit the existing permissions file, changing both `name` and the last
...
+Jenkins plugins and other components can be continuously delivered through a supported process described in [JEP-229](https://github.com/jenkinsci/jep/blob/master/jep/229/README.adoc).
+
+You can enable JEP-229 CD for your component by adding the following to your component's YAML file:
+
+```yaml
+cd:
+  enabled: true
+```
+
+**NOTE:** As of July 2022, this does not work for multimodule projects deploying multiple components, as these would have multiple YAML files in this repository.
```
dwnusbaum, line 90: Really? I am pretty sure that https://github.com/jenkinsci/pipeline-model-definition-plugin gets released via CD. See https://github.com/jenkins-infra/repository-permissions-updater/pull/2324.

timja: as does JCasC and a few others

daniel-beck: Huh, interesting. I didn't think that worked. #1747 PR comment indicates it wouldn't:

> In fact, the first item above is made worse here, this might create tokens that can only upload a subset of artifacts in a repo.

That doesn't seem to be the case looking at the code, as the per-repo group (to which the per-repo token is added) is part of all individual permission targets for each listed artifact. I guess I might have considered the edge case when not every single artifact in a GH repo declares the correct and identical `github` value? I don't remember.

Anyway, I'm happy to delete this (and even happier to learn that past me might actually have considered this problem).

```suggestion
```

PR #2693, opened 2022-07-30
Note that JEP-229 may grant release permission to unexpected users

daniel-beck:


README.md
```diff
@@ -87,11 +87,11 @@ cd:
   enabled: true
 ```
 
-
 **IMPORTANT:**
 When using JEP-229 CD, every committer to your repository can create new releases by merging pull requests.
 As a result, the list of maintainer accounts maintained in your plugin's YAML file is no longer the single reference on who can publish new releases.
-The users listed there still serve as the contacts for security issues and plugin/component governance questions.
+Be sure to check [which users have commit access](https://www.jenkins.io/doc/developer/publishing/source-code-hosting/) to your repository and remove any that are unexpected, as well as any [deploy keys](https://docs.github.com/en/developers/overview/managing-deploy-keys)
```
daniel-beck, line 93:
```suggestion
Be sure to check [which users have commit access](https://www.jenkins.io/doc/developer/publishing/source-code-hosting/) to your repository and remove any that are unexpected before enabling CD, as well as any unexpected [deploy keys](https://docs.github.com/en/developers/overview/managing-deploy-keys).
```

timja (commented):
I'm not sure if this is the most appropriate place to document this?

Shouldn't it be on jenkins.io instead?

i.e. would people actually see this.

daniel-beck:
> I'm not sure if this is the most appropriate place to document this?
> 
> Shouldn't it be on jenkins.io instead?
> 
> i.e. would people actually see this.

timja (approved):

PR #2716, opened 2022-08-12
Use new Artifactory access tokens API

daniel-beck:
Supersedes #2683. Subsumes #1926.

The classic Artifactory tokens API (deprecated since fall 2021) has a limit for usernames of 58 chars, which is not enough for some wordy plugin IDs. In #2683 we discussed how best to deal with this, with suggestions like hashing the name (making reconstruction difficult), or shortening it with various schemes (introducing inconsistencies or changing user names for existing components).

Given the similarity of the deprecated and new APIs, it seemed easiest to switch to the new API with a documented user name limit of 255 chars. An annoying side effect is that endpoint only supports Bearer authentication with tokens, not Basic with username/(encrypted) password as before, so this changes the Jenkinsfile and how the script authenticates requests accordingly. I created the new token credential on trusted.ci.

lazysodium-java, used to encrypt the secrets for GitHub, does not work on ARM64 (on which I test) in version 5.1.1, and 5.1.2 is not released to repo1, so this also includes #1926 to remove the library.

So far I've confirmed this creates tokens, and attaches secrets to GH repos. Ideally someone has a less used plugin with CD enabled that could use a release, then I'll point my local environment at that plugin's repo for an end-to-end test before we merge.


jetersen:
@daniel-beck https://github.com/jenkinsci/jenkins-infra-test-plugin ?

timja:
As Joseph said that is a test plugin you can do whatever you want with it, mostly used for testing JEP-229 but also just plugin releases in general without bothering users

timja (approved):
If it works

jetersen (commented):
Overall LGTM, lets see how the end to end test goes 👏

daniel-beck:
https://github.com/jenkinsci/jenkins-infra-test-plugin/runs/7802603662?check_suite_focus=true looks good? It uploaded https://repo.jenkins-ci.org/artifactory/releases/io/jenkins/plugins/jenkins-infra-test/84.vfcda_97350891/ two minutes after I refreshed the tokens in https://github.com/jenkinsci/jenkins-infra-test-plugin/settings/secrets/actions; so it used the ones I submitted from my local environment.

timja:
🚢 ?

jetersen:
You should be able to verify it here, by looking at the deployed by.

https://repo.jenkins-ci.org/ui/repos/tree/General/releases/io/jenkins/plugins/jenkins-infra-test/87.v2c52051599ee/jenkins-infra-test-87.v2c52051599ee.hpi

You could temporarily the token name to verify that it uses the token you generated.

daniel-beck:
> You could temporarily the token name to verify that it uses the token you generated.

Yep, which is my point above, I forgot to do that for 84.v… (and fought against JEP-229 validation since then). However, _that_ release now has it, see the "dev" prefix.

jetersen (approved):
I see the dev prefix in 87! So all good.

daniel-beck:
> 🚢 ?

If you don't want to wait for more reviews and are happy with the tests done 🤷

jetersen:
🚢

We can always fix it 👏

timja:
> > 🚢 ?
> 
> If you don't want to wait for more reviews and are happy with the tests done 🤷

The main change is https://github.com/jenkins-infra/repository-permissions-updater/pull/1926 which has had over a year to have a review.

Testing looks good, we can always either fix-forward or rollback depending on issues anyway

jetersen:
Please edit the PR description to link the issue fixed:

fixes #2716

jetersen:
@daniel-beck Perhaps worth writing a short script that cleans up old MAVEN_USERNAME and MAVEN_PASSWORD in the repos?

src/main/java/io/jenkins/infra/repository_permissions_updater/CryptoUtil.java
```diff
@@ -1,16 +1,13 @@
...
 import edu.umd.cs.findbugs.annotations.NonNull;
+import org.kocakosm.jblake2.Blake2b;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Base64;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+
+import static com.neilalexander.jnacl.crypto.curve25519xsalsa20poly1305.*;
```
jglick, line 10: :exploding_head:

jglick (commented):
Thanks!

PR #2719, opened 2022-08-12
Fix XML tag mentions in hosting checker

daniel-beck:
Because this output is clearly broken:

> Screenshot 2022-08-12 at 21 12 43

Untested except for in this PR comment:

#### Unescaped:

<licenses>
<groupId>

#### Escaped:

&lt;licenses&gt;
&lt;groupId&gt;


MarkEWaite (approved):

timja (approved):
backticks would also work as well I assume but whatever works, thanks!

PR #2891, opened 2022-10-21
Automatically detects plugins up for adoption

alecharp:
The tasks to mark a plugin up for adoption is manual. 
However, when the `developers` list is empty, no one has the permission to push a new release to Artifactory.
It means that no one (admins and security officers don't count) can push a new bug fix to users.

Out of fairness for the users, I think it would be ok to mark any plugin without any `developers` as up for adoption. 

This is of course, up for debate.

### DIFF
--- a/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
+++ b/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
@@ -123,6 +123,7 @@ class ArtifactoryPermissionsUpdater {
         Map<String, List> issueTrackersByPlugin = new TreeMap()
         Map<String, List<Definition>> cdEnabledComponentsByGitHub = new TreeMap<>()
         Map<String, List<String>> maintainersByComponent = new HashMap<>()
+        List<String> repositoriesUpForAdoption = new ArrayList<>()
 
         yamlSourceDirectory.eachFile { file ->
             if (!file.name.endsWith('.yml')) {
@@ -230,6 +231,11 @@ class ArtifactoryPermissionsUpdater {
                 principals {
                     if (definition.developers.length == 0) {
                         users [:]
+                        if (definition.github) {
+                            repositoriesUpForAdoption.add(definition.github)
+                        } else {
+                            LOGGER.log(Level.WARNING, "${definition.name} has no GitHub details, will not be able to add adoption label")
+                        }
                     } else {
                         users definition.developers.collectEntries { developer ->
                             def existsInArtifactory = KnownUsers.existsInArtifactory(developer)
@@ -316,6 +322,10 @@ class ArtifactoryPermissionsUpdater {
         def maintainers = new JsonBuilder()
         maintainers(maintainersByComponent)
         new File(apiOutputDir, 'maintainers.index.json').text = maintainers.toPrettyString()
+
+        def adoptionRepos = new JsonBuilder()
+        adoptionRepos(repositoriesUpForAdoption)
+        new File(apiOutputDir, 'adoption.index.json').text = adoptionRepos.toPrettyString()
     }
 
     // TODO It's a really weird decision to have this in the otherwise invocation agnostic standalone tool
@@ -435,6 +445,18 @@ class ArtifactoryPermissionsUpdater {
         }
     }
 
+    private static void addAdoptionTopicOnRepositories(File githubReposForAdoptionIndex) {
+        def repos = new JsonSlurper().parse(githubReposForAdoptionIndex);
+        def topic = "adopt-this-plugin"
+        repos.each { repo ->
+            if (DRY_RUN_MODE) {
+                LOGGER.log(Level.INFO, "Skipped adding adoption topic for GitHub repo: '${repo}'")
+                return
+            }
+            GitHubAPI.getInstance().addTopicToRepository((String) repo, topic)
+        }
+    }
+
     static void main(String[] args) throws IOException {
         for (Handler h : Logger.getLogger("").getHandlers()) {
             if (h instanceof ConsoleHandler) {
@@ -469,6 +491,10 @@ class ArtifactoryPermissionsUpdater {
          * For all CD-enabled GitHub repositories, obtain a token from Artifactory and attach it to a GH repo as secret.
          */
         generateTokens(new File(ARTIFACTORY_API_DIR, "cd.index.json"))
+        /*
+         * For all plugins without developers, mark them as up for adoption
+         */
+        addAdoptionTopicOnRepositories(new File(ARTIFACTORY_API_DIR, "adoption.index.json"))
     }
 
     private static final Logger LOGGER = Logger.getLogger(ArtifactoryPermissionsUpdater.class.name)
--- a/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
+++ b/src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
@@ -234,7 +234,7 @@ class ArtifactoryPermissionsUpdater {
                         if (definition.github) {
                             repositoriesUpForAdoption.add(definition.github)
                         } else {
-                            LOGGER.log(Level.WARNING, "${definition.name} has no GitHub details, will not be able to add adoption label")
+                            LOGGER.log(Level.WARNING, "${definition.name} has no GitHub details, will not be able to add adoption topic")
                         }
                     } else {
                         users definition.developers.collectEntries { developer ->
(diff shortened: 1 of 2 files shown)


src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
```diff
@@ -89,19 +92,46 @@ abstract class GitHubAPI {
...
         void createOrUpdateRepositorySecret(String name, String encryptedSecret, String repositoryName, String keyId) {
             LOGGER.log(Level.INFO, "Create/update the secret ${name} for ${repositoryName} encrypted with key ${keyId}")
-            URL url = new URL("https://api.github.com/repos/${repositoryName}/actions/secrets/${name}")
-            HttpURLConnection conn = (HttpURLConnection) url.openConnection()
-
-            // The GitHub API doesn't do an auth challenge
-            conn.setRequestProperty('Authorization', 'Basic ' + Base64.getEncoder().encodeToString((System.getenv("GITHUB_USERNAME") + ':' + System.getenv("GITHUB_TOKEN")).getBytes(StandardCharsets.UTF_8)))
-            conn.setRequestProperty('Accept', 'application/vnd.github.v3+json')
-            conn.setRequestMethod('PUT')
+            HttpURLConnection conn = createConnection("https://api.github.com/repos/${repositoryName}/actions/secrets/${name}", 'PUT')
```
lemeurherve, line 95:
```suggestion
            HttpURLConnection conn = githubRequest("https://api.github.com/repos/${repositoryName}/actions/secrets/${name}", 'PUT')
```

src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
```diff
@@ -72,13 +81,7 @@ abstract class GitHubAPI {
...
         GitHubPublicKey getRepositoryPublicKey(String repositoryName) {
             LOGGER.log(Level.INFO, "GET call to retrieve public key for ${repositoryName}")
-            URL url = new URL("https://api.github.com/repos/${repositoryName}/actions/secrets/public-key")
-            HttpURLConnection conn = (HttpURLConnection) url.openConnection()
-
-            // The GitHub API doesn't do an auth challenge
-            conn.setRequestProperty('Authorization', 'Basic ' + Base64.getEncoder().encodeToString((System.getenv("GITHUB_USERNAME") + ':' + System.getenv("GITHUB_TOKEN")).getBytes(StandardCharsets.UTF_8)))
-            conn.setRequestProperty('Accept', 'application/vnd.github.v3+json')
-            conn.setRequestMethod('GET')
+            HttpURLConnection conn = createConnection("https://api.github.com/repos/${repositoryName}/actions/secrets/public-key", 'GET')
```
lemeurherve, line 84:
```suggestion
            HttpURLConnection conn = githubRequest("https://api.github.com/repos/${repositoryName}/actions/secrets/public-key", 'GET')
```
It makes more sense IMO.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
```diff
@@ -89,19 +92,46 @@ abstract class GitHubAPI {
...
+            HttpURLConnection conn = createConnection("https://api.github.com/repos/${repositoryName}/actions/secrets/${name}", 'PUT')
             conn.setDoOutput(true)
             OutputStreamWriter osw = new OutputStreamWriter(conn.getOutputStream())
             osw.write('{"encrypted_value":"' + encryptedSecret + '","key_id":"' + keyId + '"}')
             osw.close()
 
             String text = conn.getInputStream().getText()
         }
+
+        private static HttpURLConnection createConnection(String url, String method) {
```
lemeurherve, line 104:
```suggestion
        private static HttpURLConnection githubRequest(String url, String method) {
```

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -230,6 +231,11 @@ class ArtifactoryPermissionsUpdater {
                 principals {
                     if (definition.developers.length == 0) {
                         users [:]
+                        if (definition.github) {
+                            repositoriesUpForAdoption.add(definition.github)
+                        } else {
+                            LOGGER.log(Level.WARNING, "${definition.name} has no GitHub details, will not be able to add adoption label")
```
lemeurherve, line 237:
```suggestion
                            LOGGER.log(Level.WARNING, "${definition.name} has no GitHub details, will not be able to add adoption topic")
```
I would also add simple quotes around the topic name in all messages. (nit)

lemeurherve (commented):
This seems a great addition to me!

Some small suggestions.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
```diff
@@ -89,19 +92,46 @@ abstract class GitHubAPI {
...
+                json.names.add(topic)
+                conn = createConnection("https://api.github.com/repos/${repositoryName}/topics", 'PUT')
+                conn.setDoOutput(true)
+                def osw = new OutputStreamWriter(conn.getOutputStream())
+                osw.write(JsonOutput.toJson(json))
+                osw.close()
+
+                conn.getInputStream().getText();
+            } else {
+                LOGGER.log(Level.INFO, "Repository ${repositoryName} already has topic ${topic}")
```
alecharp, line 133:
```suggestion
                LOGGER.log(Level.INFO, "Repository ${repositoryName} already has topic '${topic}'")
```

src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
```diff
@@ -89,19 +92,46 @@ abstract class GitHubAPI {
...
+            HttpURLConnection conn = (HttpURLConnection) new URL(url).openConnection()
+
+            conn.setRequestProperty('Authorization', 'Basic ' + Base64.getEncoder().encodeToString((System.getenv("GITHUB_USERNAME") + ':' + System.getenv("GITHUB_TOKEN")).getBytes(StandardCharsets.UTF_8)))
+            conn.setRequestProperty('Accept', 'application/vnd.github.v3+json')
+            conn.setRequestMethod(method)
+        }
+
+        @Override
+        void addTopicToRepository(String repositoryName, String topic) {
+            LOGGER.log(Level.INFO, "Add topic ${topic} on ${repositoryName}")
```
alecharp, line 114:
```suggestion
            LOGGER.log(Level.INFO, "Add topic '${topic}' on ${repositoryName}")
```

timja:
> For CD-enabled plugins ideally we would grant GH write permission to maintainers without also granting Artifactory permissions

It's currently also a fallback not removing from RPU so if trusted.ci breaks you can still release a CD plugin without having to wait till it gets fixed.

and security team uses it for security contacts...

jglick:
> so if trusted.ci breaks you can still release a CD plugin

Huh? CD releases operate via GHA. Admittedly there are various reasons why CD might be temporarily broken and you might wish to cut an emergency release manually without filing a helpdesk ticket.

timja:
> Admittedly there are various reasons why CD might be temporarily broken and you might wish to cut an emergency release manually without filing a helpdesk ticket

exactly

> Huh?

tokens aren't valid very long and require trusted.ci to be constantly updating them

alecharp:
For the record, one of the first tasks to adopt a plugin in https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/ is to remove the topic on the repository. So I didn't expect the permissions to be a problem.

> I doubt many/any of the plugins you found would be affected, but I would recommend skipping this logic when `cd.enabled` is set.

This is something I can add yes. However, a plugin can be release and still have no maintainers. An admin or a security member could theoretically merge pull requests from dependabot and force a release through CD. And yet, the plugin should be mark as up for adoption. 

This is a flag to invite a new person willing to work on a plugin to do so.

daniel-beck:
IMO this would be better as a one-off cleanup and then something just done on merge when removing the last maintainer from the list. I don't see a need for this to run every 30 or so minutes.

alecharp:
I feel like, because it's easy to forget, having it automatic, we (not me but others) won't have to think about it.

halkeye:
Why not just have have update center as reporting it as up for adoption. It already has the facilities to add labels and read the developers list and stuff, and wouldn't need an extra file to be archived/published and co summed by others.

daniel-beck:
> Why not just have have update center as reporting it as up for adoption. It already has the facilities to add labels and read the developers list and stuff, and wouldn't need an extra file to be archived/published and co summed by others.

Great idea. Will do that next weekend or before then.

alecharp:
any solution is fine to me, as long as we have one ;)

daniel-beck:
FTR I filed https://github.com/jenkins-infra/update-center2/pull/654 and https://github.com/jenkins-infra/repository-permissions-updater/issues/2911

PR #2965, opened 2022-11-17
Bump minimum parent pom version to 4.50

NotMyFault:
There are a plenty of benefits in using newer versions, than 4.0, like support for Java 17, JUnit 5 support in JenkinsRule or a lot of policy adaptions, like making a `index.jelly` mandatory.
Not to forget all the bug fixes and other feature additions.

Therefore, I'd like to move forward, and bump the minimum parent pom version to a recent release.


MarkEWaite (approved):
+1 from me.  It is awkward to have newly hosted plugins that immediately need multiple steps from the "Improve a plugin" tutorial.

PR #2994, opened 2022-11-24
Require linking to the downstream PR when enabling JEP-229

NotMyFault:
On a regular basis, I notice, that plugins enabling CD do not follow the guidelines properly.
Missing or wrong incremental format, wrong (and uncalled) modifications to the action workflow or overall modifications that break CD or publishes improper releases.

I'd like to propose that we require submitters to include a link to their PR, which prepares their plugin for JEP-229, allowing us to scan over it quickly.
Of course, this is no hard requirement or set in stone, people with a reputable history of plugins using CD properly, are more than welcome to skip this part.
The target audience are first time requestors or people who are new to our tooling; the people I often found issues in their CD setup.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,23 +7,17 @@
 # Submitter checklist for adding or changing permissions
 
 <!--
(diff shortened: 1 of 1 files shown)


halkeye (approved):

halkeye:
Longer term, a guided form thingie might be better than a template

NotMyFault:
> Longer term, a guided form thingie might be better than a template

Hopefully, we don't need a long term solution. Get us a good chunk of proper CD PRs people can refer to.

PR #3015, opened 2022-12-06
Suspend Gradle checks for new plugins

NotMyFault:
Picks up my comment from https://github.com/jenkins-infra/helpdesk/issues/3279#issuecomment-1339681950

I suggest we do no longer host new plugins using the JPI plugin, until it supports the same scope as the maven architecture. As of now, there's no built-in release functionality, no support for JEP-229 (CD) and probably a few other components, which have been added within the past years.


timja (changes requested):
this should be a discussion on the dev mailing list first

NotMyFault:
Linking https://groups.google.com/g/jenkinsci-dev/c/lHQAiEepBiw for cross reference

NotMyFault:
Overall consensus is positive, do you think we can continue with this PR?

timja (approved):

PR #3043, opened 2022-12-21
More adoption for WMI

NotMyFault:
All plugins are either labeled as up for adoption or are unmaintained.

### DIFF
--- a/permissions/plugin-downstream-ext.yml
+++ b/permissions/plugin-downstream-ext.yml
@@ -6,3 +6,5 @@ issues:
 paths:
   - "org/jenkins-ci/plugins/downstream-ext"
 developers: []
+cd:
+  enabled: true
--- a/permissions/plugin-doxygen.yml
+++ b/permissions/plugin-doxygen.yml
@@ -7,3 +7,5 @@ paths:
   - "org/jenkins-ci/plugins/doxygen"
 developers:
   - "gbois"
+cd:
+  enabled: true
--- a/permissions/plugin-jqs-monitoring.yml
+++ b/permissions/plugin-jqs-monitoring.yml
@@ -7,3 +7,5 @@ issues:
 paths:
   - "org/jenkins-ci/plugins/jqs-monitoring"
 developers: []
+cd:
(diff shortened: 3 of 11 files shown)


permissions/plugin-downstream-ext.yml
```diff
@@ -6,3 +6,5 @@ issues:
 paths:
   - "org/jenkins-ci/plugins/downstream-ext"
 developers: []
```
MarkEWaite, line 8: I think that you want to be a developer here as well
```suggestion
developers:
  - "notmyfault"
```

permissions/plugin-doxygen.yml
```diff
@@ -7,3 +7,5 @@ paths:
   - "org/jenkins-ci/plugins/doxygen"
 developers:
   - "gbois"
```
MarkEWaite, line 9: A developer here also
```suggestion
developers:
  - "gbois"
  - "notmyfault"
```

permissions/plugin-jqs-monitoring.yml
```diff
@@ -7,3 +7,5 @@ issues:
 paths:
   - "org/jenkins-ci/plugins/jqs-monitoring"
 developers: []
```
MarkEWaite, line 9: Developer here also
```suggestion
  - "org/jenkins-ci/plugins/jqs-monitoring"
developers:
  - "notmyfault"
```

MarkEWaite (approved):
Thanks for doing it.  Optional additional changes for your consideration

NotMyFault:
I left them out on purpose, given I'm planning to release via CD and only be around for a short timeframe, until the deprecated dependency is dropped/the plugin has been modernized. I won't stay as maintainer once these tasks are done, hence I decided to keep the PR noise low, by just enabling CD.
But I can add myself if you want.

> Thanks for doing it.

You're welcome :)
Plugin modernization may also be something for newbie contributors or people getting into OSS following the guides on jenkins.io, hence I'll stick to plugins that are up for adoption or have no maintainer.

PR #3140, opened 2023-02-10
Remove Jira field from JUnit plugin

NotMyFault:
ref https://issues.jenkins.io/browse/JENKINS-70600.

The jira component has been archived.

### DIFF
--- a/permissions/plugin-junit.yml
+++ b/permissions/plugin-junit.yml
@@ -2,7 +2,6 @@
 name: "junit"
 github: &gh "jenkinsci/junit-plugin"
 issues:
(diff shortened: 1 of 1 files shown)


slide:
Only a nit, it looks like other strings are using double quotes in the file, maybe stick with the same for this?

MarkEWaite (approved):

MarkEWaite:
> Only a nit, it looks like other strings are using double quotes in the file, maybe stick with the same for this?

Since this is deleting a single quoted string, I think the pull request matches with your request.

PR #3148, opened 2023-02-15
Validate diverging groupIds

NotMyFault:
Let's ensure new hosting requests always use `io.jenkins.plugins` as groupId.


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
```diff
@@ -133,8 +133,8 @@ private void checkGroupId(Model model, HashSet<VerificationMessage> hostingIssue
         try {
             String groupId = model.getGroupId();
             if(StringUtils.isNotBlank(groupId)) {
-                if(groupId.equals("org.jenkins-ci.plugins")) {
-                    hostingIssues.add(new VerificationMessage(VerificationMessage.Severity.REQUIRED, SHOULD_BE_IO_JENKINS_PLUGINS));
+                if (groupId.equals("org.jenkins-ci.plugins") || !groupId.equals("io.jenkins.plugins")) {
```
timja, line 136: this would be simpler:
```suggestion
                if (!groupId.equals("io.jenkins.plugins")) {
```

NotMyFault: true dat

timja (approved):

PR #3156, opened 2023-02-24
chore: pin GitHub actions with their SHA-1 instead of their version number

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3402

### DIFF
--- a/.github/workflows/hosting-comment-checker.yml
+++ b/.github/workflows/hosting-comment-checker.yml
@@ -1,11 +1,9 @@
 name: Hosting re-checker
-
 on:
   issue_comment:
     types:
       - created
       - edited
-
 jobs:
   hosting:
     runs-on: ubuntu-latest
--- a/.github/workflows/hosting-comment-hoster.yml
+++ b/.github/workflows/hosting-comment-hoster.yml
@@ -1,11 +1,9 @@
 name: Hosting hoster
-
 on:
   issue_comment:
     types:
       - created
       - edited
-
 jobs:
   validate:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


timja (changes requested):
can we leave actions/checkout and actions/setup-java with a pinned version instead please? to reduce dependabot pull requests I already get a million of them...

NotMyFault:
Preferably, only 3rd party actions should be pinned. Official actions administered by GH, such as `actions/`, or our actions, should stay as-is, no?

lemeurherve:
> can we leave actions/checkout and actions/setup-java with a pinned version instead please? to reduce dependabot pull requests I already get a million of them...

I'm currently updating my script to let only the version in the comment.

> Preferably, only 3rd party actions should be pinned. Official actions administered by GH, such as `actions/`, or our actions, should stay as-is, no?

I don't see why, and it would also be more complicated as I would need to exclude these modifications.

dduportal:
> can we leave actions/checkout and actions/setup-java with a pinned version instead please? to reduce dependabot pull requests I already get a million of them...

For any action, is dependabot triggered on each SHA change (e.g. on each commit on the action's source repository) or is it the same as before (e.g. triggered when a new release is available, but the SHA is updated along with the commented version)?

timja:
> > can we leave actions/checkout and actions/setup-java with a pinned version instead please? to reduce dependabot pull requests I already get a million of them...
> 
> For any action, is dependabot triggered on each SHA change (e.g. on each commit on the action's source repository) or is it the same as before (e.g. triggered when a new release is available, but the SHA is updated along with the commented version)?

No it's when it's released but we pin to e.g. version 3 rather than 3.10 as we are happy to receive minor updates to these actions.

timja (approved):

dduportal:
> No it's when it's released but we pin to e.g. version 3 rather than 3.10 as we are happy to receive minor updates to these actions.

Make sense. Looks like that if we do not "sha-ized" the `actions/` that would help (ref. https://github.com/jenkins-infra/helpdesk/issues/3402#issuecomment-1443706544).

FWIW, if we ran into a 3rd party GHA too verbose, we can always tell dependabot to ignore the patch or minor versions: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the- dependabot.yml-file#ignore

PR #3186, opened 2023-03-14
Fix miniorange user name

daniel-beck:
The actual username is all lowercase (I thought we had validation for that, but apparently not).

https://issues.jenkins.io/secure/ViewProfile.jspa?name=miniorange


timja (approved):

NotMyFault (approved):
Isn't LDAP case insensitive?

daniel-beck:
@NotMyFault 

Don't assume there are no badly written shell scripts somewhere that are 😉

PR #3203, opened 2023-03-21
chore(Jenkinsfile) retry the "Run" stage if it fails

dduportal:
# Description

Related to https://github.com/jenkins-infra/helpdesk/issues/3462#issuecomment-1477681324

This PR updates the Jenkinfile to add a `retry()` step in the `Run` stage when not in "dry run" mode.

The goal is to avoid expired RPU tokens which are valid 5 hours: as the build runs every 3 hours, any build failure usually ends in HTTP/401 for releases.

Retrying the stage `Run`, which is the most prone to fail due to GitHub API HTTP/5xx errors, is a tentative of improvement.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,7 +77,9 @@ node('maven-11') {
                         string(credentialsId: 'artifactoryAdminToken', variable: 'ARTIFACTORY_TOKEN'),
                         usernamePassword(credentialsId: 'jenkins-infra-bot-github-token', passwordVariable: 'GITHUB_TOKEN', usernameVariable: 'GITHUB_USERNAME')
                 ]) {
-                    sh 'java ' + javaArgs
+                    retry(conditions: [agent(), nonresumable()], count: 2) {
+                        sh 'java ' + javaArgs
+                    }
                 }
             }
         }


NotMyFault (approved):

daniel-beck:
Not a good solution IMO given the fairly long run time of the script (the more JEP-229, the worse). Would do this in the code instead. For the specific error encountered today, wrapping https://github.com/jenkins-infra/repository-permissions-updater/blob/master/src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy#L93-L102 in a loop with 2-3 tries, each a few seconds apart, should do it.

dduportal:
We had another issue (https://github.com/jenkins-infra/helpdesk/issues/3571) caused by an HTTP/500 error from GitHub.

Each time we have this kind of issue, we have to relaunch the RPU manually as soon as posible.


Any objection if I merge this, given it only retries a second time (no more)?


(I understand @daniel-beck comment and I agree it would be better but I do'nt have the skills neither the time to do it so I would want to start with a "retstart once to cver 80% + of the error cases)

dduportal:
I'm proceeding to merge to help the infra team in its supporting task.

I don't mind rollbacking if it is causing trouble, or if someone proposes a "native" change.

PR #3244, opened 2023-04-11
Update hosting message with last archetype buildPlugin model

lemeurherve:
# Description

Update the hosting welcome message with [the last archetype `buildPlugin` model](https://github.com/jenkinsci/archetypes/blob/master/common-files/Jenkinsfile).


timja (commented):
What about just pointing at the one in archetypes so we don't have to maintain it here? (especially in a string)

NotMyFault (approved):
Thanks!

PR #3308, opened 2023-05-13
Reword who reviews hosting requests

NotMyFault:


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
```diff
@@ -81,7 +81,8 @@ public void checkRequest(int issueID) throws IOException {
             msg.append("\nYou can re-trigger a check by editing your hosting request or by commenting `/hosting re-check`");
         } else {
             msg.append("It looks like you have everything in order for your hosting request. "
-                    + "A human volunteer will check over things that I am not able to check for "
+                    + "A member of the [Jenkins hosting team](https://www.jenkins.io/project/teams/hosting/#members-of-the-hosting-team) "
+                    + "will check over things that I am not able to check for "
```
MarkEWaite, line 85: Maybe drop that last "for"
```suggestion
                    + "will check over things that I am not able to check"
```

MarkEWaite (approved):
Optional comment for consideration.  Approved whether the comment is accepted or not

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
NotMyFault: Good catch!

PR #3260, opened 2023-04-19
Added Jenkins infrastructure account under developer section

pavlovic-ivan:
# Description

I am requesting permission to be added as an uploader for the plugin named `mstest-plugin`. The corresponding GitHub repository can be found [here](https://github.com/jenkinsci/mstest-plugin). @jetersen has instructed me to adopt the plugin and to start with this pull request, requesting permissions. This can be seen [here](https://github.com/jenkinsci/mstest-plugin/pull/16#issuecomment-1502937834). My Github account is @pavlovic-ivan that should also receive access to `jenkinsci` Github organisation.

# Submitter checklist for adding or changing permissions

### Always

- [x] Add link to plugin/component Git repository in description above

### When enabling automated releases (cd: true)

- [ ] Add a link to the pull request, which enables continuous delivery for your plugin or component.  
Follow the [documentation](https://www.jenkins.io/doc/developer/publishing/releasing-cd/) to ensure, your pull request is set up properly.

### When adding new uploaders (this includes newly created permissions files)

- [x] [Make sure to `@`mention an existing maintainer to confirm the permissions request, if applicable](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)
- [x] Use the Jenkins community (LDAP) account name in the YAML file, not the GitHub account name
- [x] Make sure to `@`mention the users being added so their GitHub account names are known if they require GitHub merge access (see below).
- [x] [All newly added users have logged in to Artifactory and Jira at least once](https://github.com/jenkins-infra/repository-permissions-updater/#requesting-permissions)

### Reviewer checklist (not for requesters!)

- [ ] Check this if newly added person also needs to be given merge permission to the GitHub repo (please @ the people/person with their GitHub username in this issue as well). If needed, it can be done using an [IRC Bot command](https://jenkins.io/projects/infrastructure/ircbot/#github-repo-management)
- [ ] Check that the `$pluginId Developers` team has `Admin` permissions while granting the access.
- [ ] In the case of plugin adoption, ensure that the Jenkins Jira default assignee is either removed or changed to the new maintainer.
- [ ] If security contacts are changed (this includes add/remove), ping the security officer (currently `@Wadeck`) in this pull request. If an email contact is changed, wait for approval from the security officer.

There are [IRC Bot commands](https://jenkins.io/projects/infrastructure/ircbot/#issue-tracker-management) for it

PR #3359, opened 2023-06-14
Unblock releases of `mstest`

yaroslavafenkin:


daniel-beck:
Originally introduced by @NotMyFault in https://github.com/jenkins-infra/repository-permissions-updater/pull/3260.

To clarify, the security team blocks releases of plugins when there are vulnerabilities introduced between the latest release and the current state of the default branch. A plugin having active security warnings is not a reason to block further releases.

NotMyFault (approved):
I was thinking cutting releases with open vulnerabilities causes more work down the road, such as updating CVE records.

Thanks for the clarification!

PR #3615, opened 2023-11-14
Block CD releases of jobConfigHistory as it has no maintainers

daniel-beck:
Amends https://github.com/jenkins-infra/repository-permissions-updater/pull/3313

Probably a useful addition to the tool instead of fixing this in data, but that's for some other time.


MarkEWaite (approved):

daniel-beck:
> Probably a useful addition to the tool instead of fixing this in data, but that's for some other time.

#3616

PR #3616, opened 2023-11-14
Support exclusive CD mode, disallow CD releases if there are no maintainers

daniel-beck:
Amends/reverts https://github.com/jenkins-infra/repository-permissions-updater/pull/3313. Obsoletes https://github.com/jenkins-infra/repository-permissions-updater/pull/3615.

Two major new features:

1. Remove CD release permissions if there are no maintainers, obsoleting PRs like https://github.com/jenkins-infra/repository-permissions-updater/pull/3313 or https://github.com/jenkins-infra/repository-permissions-updater/pull/3615.
2. Add support for _exclusive CD_. In this mode, we keep track of maintainers for governance purposes, but they don't get upload permissions.

The latter is required to support the former, as some core(ish) components and libraries have CD but no maintainers defined, and would get their CD permissions revoked. We could add maintainers, but they would then get permission to upload releases. In those cases, I just added the `@core` group to the maintainers list, as these appear to be generally collectively maintained, rather than by specific individuals (even though it might boil down to that in practice).

## Testing

Based on comparing JSON output of PR builds, permissons etc. have no differences compared to `master` except it pre-empts #3615.

Happy to do a real run in the development involving Artifactory permissions if desired, but that's so annoying to set up I'd prefer to do that only if this otherwise looks reasonable. It's also easily reverted if things go wrong unexpectedly, IMO.

CC @NotMyFault @basil as most active recent contributors to the components I added `@core` for in case that does not reflect how you see their maintenance model.

`on-hold`: I'm going to be out for a few days and will be unable to amend if needed, so would prefer this not be merged until Nov 21.

### DIFF
--- a/permissions/plugin-anything-goes-formatter.yml
+++ b/permissions/plugin-anything-goes-formatter.yml
@@ -6,6 +6,5 @@ issues:
 paths:
   - "org/jenkins-ci/plugins/anything-goes-formatter"
 developers: []
-#CD blocked for lack of maintainers
-#cd:
-#  enabled: true
+cd:
+  enabled: true
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,8 +86,7 @@ node('maven-11') {
         }
     } finally {
         stage ('Archive') {
-            archiveArtifacts 'permissions/*.yml'
-            archiveArtifacts 'json/*.json'
+            archiveArtifacts 'json/**'
             if (infra.isTrusted()) {
                 dir('json') {
                     publishReports ([ 'issues.index.json', 'maintainers.index.json' ])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,7 +86,8 @@ node('maven-11') {
         }
     } finally {
         stage ('Archive') {
-            archiveArtifacts 'json/**'
+            archiveArtifacts 'permissions/*.yml'
+            archiveArtifacts 'json/*.json'
             if (infra.isTrusted()) {
                 dir('json') {
                     publishReports ([ 'issues.index.json', 'maintainers.index.json' ])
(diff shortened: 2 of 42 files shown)


NotMyFault:
2) sounds nice

permissions/plugin-anything-goes-formatter.yml
```diff
@@ -6,6 +6,5 @@ issues:
 paths:
   - "org/jenkins-ci/plugins/anything-goes-formatter"
 developers: []
-#CD blocked for lack of maintainers
-#cd:
-#  enabled: true
+cd:
+  enabled: true
```
timja, line 10: this doesn't really seem valid to me, isn't it clearer if we error if there's no maintainers?

timja (approved):
I like the new `exclusive`.

I'm unsure on whether we should allow no maintainers and cd.enabled = true as I feel like it might cause a bit of confusion when it doesn't work and the docs don't seem to be clear that it's required.

I'm not blocking it though so I will approve it.

daniel-beck:
> the docs don't seem to be clear that it's required

Good point, added in https://github.com/jenkins-infra/repository-permissions-updater/pull/3616/commits/5544c6d52a547fab70ae2fd4692a85fdfcb3e0d3

> this doesn't really seem valid to me, isn't it clearer if we error if there's no maintainers?

Unsure. Seems like it would be an additional barrier to adding/removing maintainers to fail if that's the case. I implemented what I thought makes sense but will defer to the folks handling hosting requests at the moment.

timja (approved):

jglick (approved):
Should we turn on exclusive mode for some plugins too? Feel free to do so for anything listed as CD-enabled and having only me (`jglick`) as a maintainer.

MarkEWaite:
If you're willing to add plugins to be exclusive CD, I'd be happy to have the following plugins switch to exclusively CD:

* basic-branch-build-strategies-plugin
* elastic-axis-plugin
* embeddable-build-status-plugin
* implied-labels-plugin
* markdown-formatter-plugin
* platformlabeler-plugin
* schedule-build-plugin
* testng-plugin-plugin
* xshell-plugin

daniel-beck:
> Should we turn on exclusive mode for some plugins too?

> I'd be happy to have the following plugins switch to exclusively CD

Feels more like a followup change to me; especially since the feature is already made use of (and thereby demonstrating it's not broken) to effectively keep existing behavior here.

basil:
> Feels more like a followup change to me; especially since the feature is already made use of (and thereby demonstrating it's not broken) to effectively keep existing behavior here.

I have a slightly different perspective, which is that this PR implicitly kicked off a migration, which was never completed.

PR #3667, opened 2023-12-04
Adopt `project-inheritance`

basil:
See https://github.com/jenkins-infra/helpdesk/issues/3842. This plugin has a deployed POM that uses HTTP instead of HTTPS, so consumers (including `promoted-builds`) can't switch off the JCenter version (which JFrog has asked us to stop caching) until `project-inheritance` is republished with HTTPS.


basil:
Never mind, I can't actually adopt this because [it is not hosted in the Jenkins GitHub organization](https://github.com/i-m-c/jenkins-inheritance-plugin).

PR #3668, opened 2023-12-04
Let markewaite adopt scp plugin to deprecate it

MarkEWaite:
## Let markewaite adopt scp plugin to deprecate it

The scp plugin was last released 13 years ago.  The last commit with any code change was almost 10 years ago.

There are two open security vulnerabilities.  The plain text credentials storage security vulnerability was published over 5 years ago.  The CSRF security vulnerability was published 18 months ago.

The last successful build on ci.jenkins.io was over 3 years ago.  The plugin does not build currently.

The `scp` command is available on all the operating systems that Jenkins supports.  Unix variants include `scp` with the OpenSSH commands.  Windows includes `scp` with their port of OpenSSH.

The installation counts of the plugin have decreased over the last 12 months.  Let's deprecate it as one more indicator for users that they should switch to more portable ways of performing scp.  Deprecation does not prevent users from installing it, but gives them another reason to consider removing the plugin from their installation.

Any one of those concerns would not be enough to justify deprecation of the plugin, but considering them all together, I think the plugin should be deprecated.

# Link to GitHub repository

https://github.com/jenkinsci/scp-plugin/tree/master

# Link to PR to be merged

* https://github.com/jenkinsci/scp-plugin/pull/23

# When modifying release permission

If you are modifying the release permission of your plugin or component, fill out the following checklist:

```[tasklist]
### Release permission checklist (for submitters)
- [x] The usernames of the users added to the "developers" section in the .yml file are the same the users use to log in to [accounts.jenkins.io](https://accounts.jenkins.io/).
- [x] All users added have logged in to [Artifactory](https://repo.jenkins-ci.org/) and [Jira](https://issues.jenkins.io/) once.
```

```[tasklist]
### Reviewer checklist
- [ ] If security contacts are changed (this includes add/remove), ping the security officer (currently `@Wadeck`) in this pull request. If an email contact is changed, wait for approval from the security officer.
```

### DIFF
--- a/permissions/plugin-scp.yml
+++ b/permissions/plugin-scp.yml
@@ -6,4 +6,5 @@ issues:
   - github: *gh
 paths:
   - "org/jvnet/hudson/plugins/scp"
-developers: []
+developers:
+  - "markewaite"


MarkEWaite:
@gounthar and I skipped this plugin while working through the Java 21 upgrades, because it does not build on the primary branch and has not been updated in many years.

I encountered it again while checking the top 250 most popular plugins with Artifactory configuration changes:

* https://github.com/jenkins-infra/helpdesk/issues/3842

Deprecating the plugin does not remove it from distribution, but it makes it clear that no one is maintaining it.  Deprecation persuades some users to remove it from their systems.

basil (commented):
Is it necessary to adopt the plugin in order to deprecate it? I thought the only action item was to file an `update-center2` PR with the deprecation, but I could be wrong.

MarkEWaite:
> Is it necessary to adopt the plugin in order to deprecate it? I thought the only action item was to file an `update-center2` PR with the deprecation, but I could be wrong.

The [deprecating a plugin documentation](https://www.jenkins.io/doc/developer/plugin-governance/deprecating-or-removing-plugin/) says

> Set a deprecated label for the plugin that will be visible on plugins.jenkins.io and in the Jenkins plugin manager. This can be done in two ways:
> 
> 1. Put a deprecated topic in the plugin’s GitHub repository. If you have multiple plugins inside a single repository, it will apply to all of them. This is the preferred approach.
> 
> 2. Add a deprecated label to the plugin entry in the Update Center’s [label-definitions.properties](https://github.com/jenkins-infra/update-center2/blob/master/resources/label-definitions.properties) file. Choose this approach if the GitHub repository contains multiple plugins and only some of the plugins in the repository are to be deprecated.
> 
> Update the plugin’s documentation to explain the reason of the deprecation.

Since the preferred path is to assign a topic on the repository, adopting the plugin seemed like the easiest way to do that.

basil:
Interesting, that documentation seems suspect to me because applying the topic doesn't allow one to add a URL linking to the cause. I think it makes more sense to do this in `update-center2` where a URL can be provided explaining the cause of the deprecation. Examples:

- https://github.com/jenkins-infra/update-center2/pull/735
- https://github.com/jenkins-infra/update-center2/pull/729
- https://github.com/jenkins-infra/update-center2/pull/719

NotMyFault (approved):

NotMyFault:
Approved anyway, given your proposed README changes, it's good to reflect the deprecation state in the README too.

PR #3718, opened 2024-01-01
Bump baseline to 2.414.3

NotMyFault:
2.401.3 will vanish from the recommended baselines by next Wednesday when we choose the next baseline. I propose to move on to 2.414.3 as baseline for new plugins


MarkEWaite (approved):
Makes sense to me.  

I believe that 2.401.3 won't be removed from the ["Choosing a Jenkins version" page](https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/) until after the February 2024 release of the next LTS baseline (probably 2.240.1 or something near it).  However, I think that  this is a good change even if my recollection is correct.

NotMyFault:
> I believe that 2.401.3 won't be removed from the ["Choosing a Jenkins version" page](https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/) until after the February 2024 release of the next LTS baseline (probably 2.240.1 or something near it)

Yep, that won't happen until .1 of the next baseline is available at the update center. But thanks for approving!

PR #3735, opened 2024-01-10
Fix user name

jeromepochat:
Fix user name (was defined as company username instead of GH username)


aneveux (approved):

jeromepochat:
> jeromepochat needs to log in to Artifactory and Jira

My Artifactory/Jira user name is `jpochat` so I believe my current username in https://github.com/jenkins-infra/repository-permissions-updater/blob/master/teams/cloudbees-developers.yml is correct => closing PR

PR #3786, opened 2024-02-29
Add secret creation retrying

timja:
Fixes https://github.com/jenkins-infra/helpdesk/issues/3967

Running a build on trusted.ci to test it (https://trusted.ci.jenkins.io:1443/job/repository-permissions-updater/36473/)

Code isn't the prettiest but it really needs more refactoring:
- moving away from httpurlconnection
- upgrading or dropping groovy would be nice as well
- upgrading java version in use would be nice too

### DIFF
--- a/src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
+++ b/src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
@@ -111,7 +111,7 @@ abstract class GitHubAPI {
                 if (responseCode != HttpURLConnection.HTTP_CREATED) {
                     sleep(200)
                     if (attemptNumber == maxAttempts) {
-                        LOGGER.log(Level.INFO, "Failed to create/update secret ${name} for ${repositoryName}, response code: ${responseCode}")
+                        LOGGER.log(Level.WARNING, "Failed to create/update secret ${name} for ${repositoryName}, response code: ${responseCode}")
                         break;
                     }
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/jenkins/infra/repository_permissions_updater/GitHubAPI.groovy
```diff
@@ -90,18 +90,35 @@ abstract class GitHubAPI {
...
+                OutputStreamWriter osw = new OutputStreamWriter(conn.getOutputStream())
+                osw.write('{"encrypted_value":"' + encryptedSecret + '","key_id":"' + keyId + '"}')
+                osw.close()
+                responseCode = conn.getResponseCode()
+
+                if (responseCode != HttpURLConnection.HTTP_CREATED) {
+                    LOGGER.log(Level.INFO, "Retrying create/update secret ${name} for ${repositoryName} attempt ${attemptNumber}/${maxAttempts}")
+                    sleep(200)
+                    if (attemptNumber == maxAttempts) {
+                        break;
```
lemeurherve, line 115: Maybe add a WARNING log here?

lemeurherve (approved):
Thanks!

lemeurherve (approved):

timja:
Just running one more build, logs were rotated before I got a chance to view them

timja:
Looks like it needs a bit more work, looking at the last failing builds I see that 3 out of 4 of them failed on the getPublicKey function, and the other 1 failed on this function.

timja:
The general case is at least working, the test builds don't have any retrying, shall we ship and see how it goes?

lemeurherve (approved):
🚢

Issue #3573, opened 2023-10-17
Set GitHub homepage to plugin site when accepting a hosting request


basil:
In #3503, https://github.com/jenkinsci/redhat-dependency-analytics-plugin was created without a GitHub homepage set. Following existing practice in popular plugins, I set the GitHub homepage to https://plugins.jenkins.io/redhat-dependency-analytics/ but it would be nice if this was done automatically when accepting the hosting request.

basil:
Why was https://github.com/jenkinsci/harbor-plugin created without a link to https://plugins.jenkins.io/harbor/?

timja:
because it is not in the automation.

basil:
Should it be done manually in the meantime?

timja:
Not in my opinion, anything that is expected to be done should be in the automation.

I have no expectation for hosting team members to be doing things manually separately.

basil:
I would expect it to be done for consistency with existing hosted repositories, regardless of whether it is currently covered by automation.

timja:
> I would expect it to be done for consistency with existing hosted repositories, regardless of whether it is currently covered by automation.

You are welcome to your expectations, but it is not a requirement of the hosting team.

The requirements for what happens when hosting a plugin are codified, not in wish-lists.

basil:
> The requirements for what happens when hosting a plugin are codified

Thanks Tim, I think this should be codified as a requirement so that newly-hosted repositories are consistent with existing ones.

PR #3798, opened 2024-03-03
Set homepage as plugins.jenkins.io for new plugins and update defaults

NotMyFault:
Fixes https://github.com/jenkins-infra/repository-permissions-updater/issues/3573

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
@@ -251,6 +251,7 @@ public class Hoster {
     private static void setupRepository(GHRepository r, boolean useGHIssues) throws IOException {
         r.enableIssueTracker(useGHIssues);
         r.enableWiki(false);
+        r.setHomepage("https://plugins.jenkins.io/" + r.getName().replace("-plugin", "") + "/");
     }
 
     /**
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
@@ -252,7 +252,6 @@ public class Hoster {
         r.enableIssueTracker(useGHIssues);
         r.enableWiki(false);
         r.setHomepage("https://plugins.jenkins.io/" + r.getName().replace("-plugin", "") + "/");
-        r.setEmailServiceHook("jenkinsci-commits@googlegroups.com");
         if (!r.getDefaultBranch().equals("main")) {
             r.setDefaultBranch("main");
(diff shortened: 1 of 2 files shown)


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
```diff
@@ -251,6 +251,9 @@ boolean forkGitHub(String owner, String repo, String newName, List<String> maint
     private static void setupRepository(GHRepository r, boolean useGHIssues) throws IOException {
         r.enableIssueTracker(useGHIssues);
         r.enableWiki(false);
+        r.setHomepage("https://plugins.jenkins.io/" + r.getName().replace("-plugin", "") + "/");
+        r.setEmailServiceHook("jenkinsci-commits@googlegroups.com");
```
NotMyFault, line 255: Apparently this is very common for many plugins and core repositories, and therefore I'd propose to enable it for new plugins too, in case someone uses the mailing list for reading purposes.

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
```diff
@@ -251,6 +251,11 @@ boolean forkGitHub(String owner, String repo, String newName, List<String> maint
...
+        if (!r.getDefaultBranch().equals("main")) {
+            r.setDefaultBranch("main");
+        }
```
NotMyFault, line 258: GitHub recommends creating new repositories with `main` over master, and it's the selected default if you do that through GitHub.
I'd like to build upon that and ensure we don't introduce new degenerative terms.

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
timja: I think this is an old thing and should be done in a separate PR and checked if its useful

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
```diff
@@ -251,6 +251,11 @@ boolean forkGitHub(String owner, String repo, String newName, List<String> maint
     private static void setupRepository(GHRepository r, boolean useGHIssues) throws IOException {
         r.enableIssueTracker(useGHIssues);
         r.enableWiki(false);
+        r.setHomepage("https://plugins.jenkins.io/" + r.getName().replace("-plugin", "") + "/");
+        r.setEmailServiceHook("jenkinsci-commits@googlegroups.com");
```
NotMyFault, line 255:
```suggestion
```

timja (approved):

PR #3827, opened 2024-03-18
Update repository URL of mock plugin

NotMyFault:
cc @jglick as a maintainer, what do you think? This would match with the other slave -> agent repos.

### DIFF
--- a/permissions/plugin-mock-slave.yml
+++ b/permissions/plugin-mock-slave.yml
@@ -1,6 +1,6 @@
 ---
 name: "mock-slave"
-github: "jenkinsci/mock-slave-plugin"
+github: "jenkinsci/mock-agent-plugin"
 cd:
   enabled: true
   exclusive: true


jglick:
Would make

```xml
<gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
```

not be accurate; to avoid confusion I prefer to have the repo name align with the `artifactId`. The display name and description have already been updated so this seems unnecessary.

NotMyFault:
No strong objections either way. I just thought it would align better with the other renamed repositories.

jglick:
…which are kind of a mess, because now we have https://github.com/jenkinsci/ssh-agent-plugin (plugin name `ssh-agent`) which runs the `ssh-agent` tool (an actual agent), and https://github.com/jenkinsci/ssh-agents-plugin (plugin name `ssh-slaves`) which runs ~slaves~ agents over SSH.

No strong opinion either, just a general preference to keep repo URLs aligned with `artifactId`s which must be stable.

PR #3936, opened 2024-05-23
JiraAPI Groovy Script conversion

TheMeinerLP:
In this PR I have converted the JiraAPI script into a Java class. I have reduced code and implemented better error handling.
API and implementation are separated and can be called via Singelton. 

Overall, this PR should work simply because the classes and calls have the same names

# Link to GitHub repository

# When modifying release permission

List the GitHub usernames of the users who should have commit permissions below:
- `@username1`
- `@username2`

This is needed in order to cut releases of the plugin or component.

If you are modifying the release permission of your plugin or component, fill out the following checklist:

```[tasklist]
### Release permission checklist (for submitters)
- [ ] The usernames of the users added to the "developers" section in the .yml file are the same the users use to log in to [accounts.jenkins.io](https://accounts.jenkins.io/).
- [ ] All users added have logged in to [Artifactory](https://repo.jenkins-ci.org/) and [Jira](https://issues.jenkins.io/) once.
- [ ] I have mentioned an [existing team member](https://github.com/orgs/jenkinsci/teams) of the plugin or component team to approve this request.
```

## When enabling automated releases (cd: true)

Follow the [documentation](https://www.jenkins.io/doc/developer/publishing/releasing-cd/) to ensure, your pull request is set up properly. Don't merge it yet.  
In case of changes requested by the hosting team, an open PR facilitates future reviews, without derailing work across multiple PRs.

### Link to the PR enabling CD in your plugin

```[tasklist]
### CD checklist (for submitters)
- [ ] I have provided a link to the pull request in my plugin, which enables CD according to the documentation. 
```

```[tasklist]
### Reviewer checklist
- [ ] If security contacts are changed (this includes add/remove), ping the security officer (currently `@Wadeck`) in this pull request. If an email contact is changed, wait for approval from the security officer.
```


NotMyFault:
SpotBugs is not fine with that.

TheMeinerLP:
I will handle that

NotMyFault (approved):

gounthar (approved):
And it now builds! 👏

pom.xml
```diff
@@ -219,6 +219,30 @@
             <version>5.10.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
```
timja, line 224: I don't think you need to use mockito-inline these days, core should be fine

pom.xml
```diff
@@ -219,6 +219,30 @@
...
+            <version>5.2.0</version>
```
timja, line 225: version is specified in the parent pom
```suggestion
```

src/test/java/io/jenkins/infra/repository_permissions_updater/JiraAPITest.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @BeforeEach
+    public void reset() throws NoSuchFieldException, IllegalAccessException {
+        backup = new Properties();
+        backup.putAll(System.getProperties());
+
+        Field instance = JiraAPI.class.getDeclaredField("INSTANCE");
+        instance.setAccessible(true);
+        instance.set(null, null);
```
timja, line 70: ⛔ do not use reflection in tests for code that you control, either expose the code by marking it as public or make it package-private and let the test access it

src/test/java/io/jenkins/infra/repository_permissions_updater/JiraAPITest.java
```diff
@@ -0,0 +1,227 @@
...
+        System.setProperties(backup);
+    }
+
+    @Test
+    void testEnsureLoadedWrongBaseUrl() {
+        System.setProperty("jiraUrl", "xx://issues.jenkins.io");
+        JiraAPI.getInstance().getComponentId("FakeData");
+        assertThat(memoryAppender.contains("Retrieving components from Jira...", Level.INFO)).isTrue();
+        assertThat(memoryAppender.contains("Failed to construct Jira URL", Level.ERROR)).isTrue();
+        System.clearProperty("jiraUrl");
```
timja, line 88: looks unneeded as you are clearing it in the aftereach, if you weren't doing that this would need to be done in a finally block so-as if the test failed it didn't impact later tests

```suggestion
```

timja (changes requested):
Can you please remove all tests that are checking log messages? Test code not the contents of logging. 

so e.g. `isUserPresent` returns true or false based on different responses from Jira using something like wiremock.

src/test/java/io/jenkins/infra/repository_permissions_updater/JiraAPITest.java
```diff
@@ -0,0 +1,223 @@
...
+import static org.mockito.Mockito.when;
+class JiraAPITest {
+
+    private static MemoryAppender memoryAppender;
+    private static HttpUrlStreamHandler httpUrlStreamHandler;
+
+    private static final Supplier<String> JIRA_URL = () -> System.getProperty("jiraUrl", "https://issues.jenkins.io");
+    private static final  Supplier<String> JIRA_USE_URL = () -> JIRA_URL.get() + "/rest/api/2/user";
+    private static final  Supplier<String> JIRA_USER_QUERY = () -> JIRA_USE_URL.get() + "?username=%s";
+    private static final Supplier<String> JIRA_COMPONENTS_URL = () -> JIRA_URL.get() + "/rest/api/2/project/JENKINS/components";
```
timja, line 41: Why are these using suppliers rather than just static strings?

TheMeinerLP: Otherwise many tests would fail because the URLs would then only be created once and thus errors would occur in the different test cases

timja: can you delete the test cases I asked about in https://github.com/jenkins-infra/repository-permissions-updater/pull/3936#pullrequestreview-2113190695 and then potentially you won't need this.

Otherwise you could use dependency injection instead by adding the url in the constructor which would be cleaner

timja:
> I would like to keep the tests to simply check intended behaviour. There is no point in just checking return values.

You could refactor to result types, e.g. an enum for some of the validation which would result in cleaner code.

TheMeinerLP:
The goal of this PR is the conversion from groovy to java and improvement of the code in minimal.
And this is related to GSoC

timja:
> The goal of this PR is the conversion from groovy to java and improvement of the code in minimal. And this is related to GSoC

I understand that but the tests that you've added are fragile and in my opinion its not a great practice to test specific log messages unless they are an external interface to another system.

TheMeinerLP:
In what sense fragile ?
Because if I now add Enum and more, that goes against the goal of what the PR should do and what is still planned

TheMeinerLP:
@timja Based on Alex's feedback I will add enum classes for the return values and use optionals to better catch the errors and add tests. Would you be happy with that?

timja:
> @timja Based on Alex's feedback I will add enum classes for the return values and use optionals to better catch the errors and add tests. Would you be happy with that?

Looking at the code, there's one 'validation' error for the username regex.
The rest look like exceptions.
None of them are actually anything to do with the user being present in Jira and reporting false isn't correct.

If you return an exception instead that would be better and clearer.

I don't know where returning optionals would be useful.

TheMeinerLP:
Okay, understood, I'll implement it that way.

timja (approved):
Thanks!

PR #3937, opened 2024-05-23
GithubAPI Groovy Script conversion

TheMeinerLP:
In this PR I have converted the GithubAPI script into a Java class. I have reduced code and implemented better error handling.
API and implementation are separated and can be called via Singelton. 

Overall, this PR should work simply because the classes and calls have the same names


TheMeinerLP:
Waiting for #3936 get merged

TheMeinerLP:
Tests needs be added

timja (commented):
I think the code would be simplified if try catches weren't added everywhere, apart from that it would be good to use brand names appropriately throughout.

It is GitHub, not Github or github

src/main/java/io/jenkins/infra/repository_permissions_updater/GitHubImpl.java
```diff
@@ -0,0 +1,171 @@
...
+            conn.setRequestProperty("Authorization", String.format(GITHUB_BASIC_AUTH_HEADER, GITHUB_BASIC_AUTH_VALUE));
+            conn.setRequestProperty("Accept", "application/vnd.github.v3+json");
+            try {
+                conn.setRequestMethod("GET");
+            } catch (ProtocolException e) {
+                throw new IOException("Protocol error for github", e);
+            }
+            try {
+                conn.connect();
+            } catch (IOException e) {
```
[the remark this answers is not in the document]
TheMeinerLP: Roger that

src/test/java/io/jenkins/infra/repository_permissions_updater/helper/URLHelper.java
```diff
@@ -0,0 +1,66 @@
...
+import java.net.URL;
+import java.net.URLStreamHandlerFactory;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * This class manage the handling of fake url for URL creation
+ */
+public sealed interface URLHelper permits URLHelper.URLHelperImpl {
```
timja, line 12: this class and a bunch of other mocks could be avoided by using https://wiremock.org/ instead of mocking the inputstream responses.

Wiremock would also mean less re-work on tests required if someone gets around to modernising the http code.

Not a blocker just a suggestion if you have time (I'll merge this now, but if you want to do a follow up that would be appreciated).

timja (approved):

src/test/java/io/jenkins/infra/repository_permissions_updater/helper/URLHelper.java
TheMeinerLP: I would rather do it as downstream PR

PR #4023, opened 2024-07-27
Improved code readbility for hosting checks

TheMeinerLP:
This PR cleans up code and changes existing code to Java 21 functions to make it more readable. Code optimizations and readability improvements were attempted. 

Unfortunately I am not able to test this because I am still working my way through it


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/verify/HostingFieldVerifierConsumer.java
```diff
@@ -0,0 +1,74 @@
...
+public final class HostingFieldVerifierConsumer implements VerifierConsumer {
+    @Override
+    public void accept(HostingRequest issue, HashSet<VerificationMessage> hostingIssues) {
+        String forkTo = issue.newRepoName();
+        String forkFrom = issue.repositoryUrl();
+        List<String> users = issue.githubUsers();
+
+        // check list of users
+        if (users.isEmpty()) {
+            hostingIssues.add(new VerificationMessage(VerificationMessage.Severity.REQUIRED, HostingConfig.RESOURCE_BUNDLE.getString("MISSING_LIST_OF_USERS")));
```
timja, line 22: what's the reasoning for this? we aren't likely to internationalise these so imo its just adding a layer of indirection.

Also I suspect a number of the messages are missing escaping for e.g. ' in a number of places its adding complexity

TheMeinerLP: The boiler code that may affect the readability of other developers, especially the messages, was simply moved to a file. With Intellij I can jump from string to resource bundle and back again, the complexity is reduced in this case with the code.

Issue #4024, opened 2024-07-28
Publication of Railflow for TestRail plugin


tharindud:
### Repository URL

https://github.com/railflow/railflow_jenkins_plugin

### New Repository Name

railflow-testrail-plugin

### Description

Railflow provides a set of command line tools for integrating testing tools and resulting reports with TestRail.

This plugin allows you to easily integrate Jenkins testing jobs with TestRail and automatically export all testing reports to TestRail using the Railflow CLI tools. Aggregating result from all your Jenkins jobs into TestRail allows teams to look at test trends, auto-assign failures via Railflow automation, create link between Jenkins and TestRail, and much more. 

### GitHub users to have commit permission

railflow
sergey-oplavin
tharindud

### Jenkins project users to have release permission

railflow

### Issue tracker

Jira

jenkins-cert-app:
<b>Security audit, information and commands</b>

The security team is auditing all the hosting requests, to ensure a better security by default.

This message informs you that a [Jenkins Security Scan](https://www.jenkins.io/redirect/jenkins-security-scan/) was triggered on your repository.
It takes ~10 minutes to complete.

<details><summary>Commands</summary>

The bot will parse all comments, and it will check if any line start with a command.

Security team only:
<ul>
	<li><code>/audit-ok</code> => the audit is complete, the hosting can continue :tada:.</li>
	<li><code>/audit-skip</code> => the audit is not necessary, the hosting can continue :tada:.</li>
	<li><code>/audit-findings</code> => the audit reveals some issues that require corrections :pencil2:.</li>
</ul>

Anyone:
<ul>
	<li><code>/request-security-scan</code> => the findings from the <a href="https://www.jenkins.io/redirect/jenkins-security-scan/" rel="nofollow">Jenkins Security Scan</a> were corrected, this command will re-scan your repository :mag:.</li>
	<li><code>/audit-review</code> => the findings from the audit were corrected, this command will ping the security team to review the findings :eyes:.
	It's only applicable when the previous audit required changes.</li>
</ul>

<i>Only one command can be requested per comment.</i>

</details>

<i>(automatically generated message, version: 1.29.6)</i>

jenkins-cert-app:
:x: Jenkins Security Scan failed.
The Security team was notified about this.

tharindud:
The requested change to replace prototype syntax with fetch API has been done. Changes have been pushed to master branch.

tharindud:
[...]

PR #4049, opened 2024-08-26
add a JellyVerifier for inline javascript and style tags

### LINKED ISSUE #3770: Scan for inline javascript in jelly files

I would be good if we prevent that in new plugins jelly files contain inline javascript in any kind, be it as script tags that is not loading a script, any of the onaction attributes or evaluations like in checkUrl. Maybe make this part of the security scanner so that also existing plugins get an alert when using inline javascript.

### Upstream changes

_No response_

mawinter69:
The JellyVerifier checks for several kinds of inline javascript and style usages which are not CSP compliant.
Following checks are implement
- inline style tags
- script tags without `src` attribute that are not of type `application/json`
- `onclick` attribute in namespaced element (this is an indicator that the deprecated onclick attribute is used that a few jellies in core support, e.g. f:radio), logged as warning
- all attributes starting with `on` in tags without a namespace (Those are typically html elements), logged as warning
- usage of `checkUrl` without `checkDependsOn`

Potentially some check might produce false positives.

Tested against issue 3310 which currently uses inline script and style.
Tested against 4024 it found usages of `onclick` attributes

Added unit tests that verifies all checks.

fixes #3770

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/JellyVerifier.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/JellyVerifier.java
@@ -1,0 +1,141 @@
+package io.jenkins.infra.repository_permissions_updater.hosting;
+
+import static java.util.regex.Pattern.CASE_INSENSITIVE;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.xml.XMLConstants;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import org.apache.commons.lang3.StringUtils;
+import org.kohsuke.github.GHContent;
+import org.kohsuke.github.GHContentSearchBuilder;
+import org.kohsuke.github.GitHub;
+import org.kohsuke.github.PagedSearchIterable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.w3c.dom.Attr;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+public class JellyVerifier implements Verifier {
+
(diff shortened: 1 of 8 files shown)


mawinter69:
@timja @NotMyFault

mawinter69:
I think once we have core and majority of plugins CSP compliant we should consider moving some or all of these checks to a maven plugin to ensure that nobody accidently reintroduces inline stuff

timja:
> I think once we have core and majority of plugins CSP compliant we should consider moving some or all of these checks to a maven plugin to ensure that nobody accidently reintroduces inline stuff

This is where the existing ones are I believe: https://github.com/jenkinsci/jenkins-test-harness/blob/b09ba623b3479730c0c340b9077f7251cc97c927/src/main/java/org/jvnet/hudson/test/JellyTestSuiteBuilder.java

timja (approved):
Thanks! Tests looks great

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/JellyVerifier.java
```diff
@@ -0,0 +1,141 @@
...
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+public class JellyVerifier implements Verifier {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(JellyVerifier.class);
+    private static final String INLINE_STYLE = "The jelly file %s contains an inline `<style>` tag";
+    private static final String INLINE_SCRIPT = "The jelly file %s contains a `<script>` tag with inline javascript.";
+    private static final String INLINE_SCRIPT_METHOD = "The jelly file %s potentially uses inline javascript attribute `%s`";
```
timja, line 36: Could these file paths be a link to GitHub?

timja (commented):
I wonder whether https://github.com/jenkins-infra/repository-permissions-updater/blob/a8a605966dee49e6ba09a6541fedc621b0ef372b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java#L97 needs adapting to only add `needs-fix` on errors.

see https://github.com/jenkins-infra/repository-permissions-updater/issues/4024#issuecomment-2309744925

currently it would add needs-fix if only an info was added. 

Generally all the hosting issues that appear are errors so we haven't really came across this, I think the warnings are more to do with how the bot ran e.g. can't find a gradle file is something wrong with the project setup

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/JellyVerifier.java
mawinter69: yes

PR #4081, opened 2024-09-09
Remove unneeded code

timja:
Not really sure why support for both component ID and component name was implemented in https://github.com/jenkins-infra/repository-permissions-updater/pull/1806

Component name was only used in one plugin in this PR https://github.com/jenkins-infra/repository-permissions-updater/pull/2111

This removes quite a bit of code that appears to be not needed at all

I've diffed the issues file and the only difference is the `plugin-repository` file I changed


src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -120,8 +120,8 @@ class ArtifactoryPermissionsUpdater {
 
         Map<String, Set<TeamDefinition>> teamsByName = loadTeams()
 
-        Map<String, Set<String>> pathsByGithub = new TreeMap()
-        Map<String, List> issueTrackersByPlugin = new TreeMap()
+        Map<String, Set<String>> pathsByGithub = new TreeMap<>()
+        Map<String, List> issueTrackersByPlugin = new TreeMap<>()
```
timja, line 124: diamond braces were missing causing type mismatch and IDE warnings

src/main/java/io/jenkins/infra/repository_permissions_updater/Definition.java
```diff
@@ -65,44 +63,29 @@ public boolean isGitHubIssues() {
         }
 
         @SuppressFBWarnings(value = "NP_UNWRITTEN_PUBLIC_OR_PROTECTED_FIELD",
-                            justification = "All calls are guarded by jira null check in isJira()")
-        private String loadComponentId(JiraComponentSource source) throws IOException {
-            String jiraComponentId = jira;
+                justification = "All calls are guarded by jira null check in isJira()")
+        private void assertJiraIdFormatValid() {
```
timja, line 67: now just checks that someone has entered a component id

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
```diff
@@ -243,7 +243,7 @@ class ArtifactoryPermissionsUpdater {
                         if (!definition.cd?.exclusive) {
                             users definition.developers.collectEntries { developer ->
                                 def existsInArtifactory = KnownUsers.existsInArtifactory(developer)
-                                def existsInJira = KnownUsers.existsInJira(developer) || JiraAPI.getInstance().isUserPresent(developer)
```
timja, line 246: I can't see why this would be needed, we use and trust the published report just like we do for artifactory.

This code was dead because it never gets executed due to everyone being in the report

daniel-beck:
> Not really sure why support for both component ID and component name was implemented in https://github.com/jenkins-infra/repository-permissions-updater/pull/1806

It's been a while, but I think it was because component ID discovery is pretty poor through the UI (which is my guess why you linked to the API instead), and these files should be accessible to edit IMO. Realistically, few people switch from GitHub to Jira, and nobody but admins can do so without assistance, so it's not a real use case.

timja:
> > Not really sure why support for both component ID and component name was implemented in #1806
> 
> It's been a while, but I think it was because component ID discovery is pretty poor through the UI (which is my guess why you linked to the API instead), and these files should be accessible to edit IMO. Realistically, few people switch from GitHub to Jira, and nobody but admins can do so without assistance, so it's not a real use case.

Yes you can get it in the UI if you remove and add the issue to a component then it shows in the history, only way I could see, but API is pretty easy to use here and as you say almost no one switches from GitHub to Jira.

src/main/groovy/io/jenkins/infra/repository_permissions_updater/ArtifactoryPermissionsUpdater.groovy
daniel-beck: > I can't see why this would be needed

Chesterton's fence.

https://github.com/jenkins-infra/repository-permissions-updater/pull/1856 / https://issues.jenkins.io/browse/INFRA-2835 explains it. I think that pre-dated corresponding hosting checks.

timja: Yes the hosting checks were added to this repository beginning of 2022 and only ever checked the report.

From what I can tell this is no longer needed, the report is now a reference and has been used as a requirement of hosting for the last couple of years without any issues.

PR #4367, opened 2025-03-17
send message to request on failure and improve retry on repo rename

mawinter69:
When a hosting requwest fails it will now send a failure message to the corresponding issue request.

Improved the error handling for renaming the repo. Before it was failing on the first retry and not actually retrying 5 times.

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
@@ -68,25 +68,31 @@ public class Hoster {
             String forkTo = hostingRequest.getNewRepoName();
 
             if (StringUtils.isBlank(forkFrom) || StringUtils.isBlank(forkTo) || users.isEmpty()) {
-                LOGGER.info("Could not retrieve information (or information does not exist) from the Hosting request");
+                String msg = "Could not retrieve information (or information does not exist) from the Hosting request";
+                LOGGER.info(msg);
+                reportHostingFailure(issueID, msg);
                 return;
             }
 
             // Parse forkFrom in order to determine original repo owner and repo name
             Matcher m = Pattern.compile("(?:https://github\\.com/)?(\\S+)/(\\S+)", CASE_INSENSITIVE).matcher(forkFrom);
             if (m.matches()) {
-                if (!forkGitHub(m.group(1), m.group(2), forkTo, users, issueTrackerChoice == IssueTracker.GITHUB)) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
```diff
@@ -106,7 +112,10 @@ private void run(int issueID) {
 
             String prUrl = createUploadPermissionPR(issueID, forkTo, users, hostingRequest.getJenkinsProjectUsers(), issueTrackerChoice == IssueTracker.GITHUB, componentId);
             if (StringUtils.isBlank(prUrl)) {
-                LOGGER.error("Could not create upload permission pull request");
+                String msg = "Could not create upload permission pull request";
+                LOGGER.error(msg);
+                reportHostingFailure(issueID, msg);
+                return;
```
mawinter69, line 118: Here it was not returning, so it would send the success message even when the creation of the uploadpermission PR failed.

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
```diff
@@ -210,17 +237,23 @@ boolean forkGitHub(String owner, String repo, String newName, List<String> maint
...
                 for (int i = 0; i < 5; i++) {
                     try {
-                        r.renameTo(newName);
+                        renameResult = renameRepository(r, newName);
                         break;
                     } catch (HttpException e) {
+                        LOGGER.warn("Failed to rename repository from " + repo + " to " +  newName, e);
                         if (e.getResponseCode() == 422) {
-                            Thread.sleep(1000);
-                            renameRepository(r, newName);
```
mawinter69, line 220: It was failing here in the last 2 hosting requests with another IOException which is not catched making the loop more or less useless.

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/Hoster.java
```diff
@@ -210,17 +237,23 @@ boolean forkGitHub(String owner, String repo, String newName, List<String> maint
...
+                            Thread.sleep(2000);
```
mawinter69, line 248: I have the impression that github somehow claims the repo is forked and available but when you try to rename the fork operation is still in progress. So give the rename more time.

alecharp:
This looks adequate to me, at least to mitigate the problem or have better understanding of it.

timja (approved):

Issue #4133, opened 2024-10-14
Document / automate common checks


timja:
### What feature do you want to see added?

TODO

timja:
cc @mawinter69 / @NotMyFault / @alecharp if you've got any to add to the list.

alecharp:
I think checkin the `Jenkinsfile` and the `src/main/resources/index.jelly`, the usage of BOM if any dependencies are pretty solid.
Your list looks good to me for now (I cannot think of anything else).

timja:
thanks added both

Brijeshthummar02:
@timja as there is no CONTRIBUTING.md file to the repository we should also add it.

timja:
Sure this issue is more dedicated to the checks for hosting requests not general contributions though.

mawinter69:
for the UI topic about controls, maybe add that plugin should be checked with dark-theme plugin

timja:
added

yashpal2104:
Can I work on this issue?

timja:
No it’s not beginner friendly

timja:
@mawinter69 I've checked off in the top list what I think has been done.

----

Do we think its worth making this easily runnable so people can do it as a pre-submit sort of thing so they arrive with a ready to go hosting request?

mawinter69:
I think the checks that I've implemented could also be added to the injected checks done during building a plugin (extending what I started in https://github.com/jenkinsci/jenkins-test-harness/pull/868 for the inline javascript checks)
Or we create from here a jar that people can run on their source tree and performs the checks. Would require to remove the things for interacting with github, an ever more stripped down debug mode just taking the source repo and printing it to stdout

beri04:
Hey, I’ve been exploring HostingChecker and the verifier classes. I’d like to work on automating the check for <maintainers> tag in pom.xml as part of this feature request. Would that be fine to start with?

PR #4374, opened 2025-03-22
add new checks to the Hosting Checker

### LINKED ISSUE #2310: Validate gitignore present in hosting checker

There's been a few plugins like this, mostly because of a bug in maven archetypes but we should statically check for this

### Upstream changes

_No response_

### LINKED ISSUE #2311: Validate `target`, `work` and `build` are not checked in to Git

Adapt hosting checker to stop people checking in generated directories that aren't supposed to be.

### Upstream changes

_No response_

mawinter69:
Addressing some of points mentioned in #4133 and some other improvements

- fix bug with the new jenkins.baseline property. The maven Model is not replacing properties, so the jenkins.version would contain `${jenkins.baseline}` and the version check is skipped.
- checks that there is no developers section in the pom
- checks that when plugins are imported that a dependencyManagement section is defined, logged as warning
- checks that when a dependencyManagement section is defined, plugin imports don't have a version (might not be 100% accurate if a plugin import is not in the bom, so only logged as info)
- when a dependencyManagement is present and contains a bom check that the version in the artifactId is in sync with the jenkins.version and that the jenkins.baseline property is used.
- checks for the existence of `Jenkinsfile` in the repo and that it contains a line that starts with `buildPlugin(`
- checks for the existence of a jenkins-security-scan yaml file in .github/workflows
- checks for the existence of a CODEOWNERS file and that it contains the expected line
- checks for the existence of `.gitignore`  and that it excludes `target` and `work`
- checks that `target` and `work` have never been checked in to the repo
- Increase minimum Jenkins version to 2.452.4
- Increase minimum parent pom to 4.88

tested against some open hosting requests and verified that the checks work as expected

fixes #2310 
fixes #2311

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
@@ -27,7 +27,7 @@ public class HostingChecker {
 
     public static final String INVALID_FORK_FROM = "Repository URL '%s' is not a valid GitHub repository (check that you do not have .git at the end, GitHub API doesn't support this).";
 
-    public static final Version LOWEST_JENKINS_VERSION = new Version(2, 452, 4);
+    public static final Version LOWEST_JENKINS_VERSION = new Version(2, 479, 3);
 
     public static void main(String[] args) throws IOException {
         new HostingChecker().checkRequest(Integer.parseInt(args[0]));
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
@@ -27,7 +27,7 @@ public class HostingChecker {
 
     public static final String INVALID_FORK_FROM = "Repository URL '%s' is not a valid GitHub repository (check that you do not have .git at the end, GitHub API doesn't support this).";
 
-    public static final Version LOWEST_JENKINS_VERSION = new Version(2, 440, 3);
+    public static final Version LOWEST_JENKINS_VERSION = new Version(2, 452, 4);
 
     public static void main(String[] args) throws IOException {
         new HostingChecker().checkRequest(Integer.parseInt(args[0]));
@@ -45,6 +45,7 @@ public class HostingChecker {
         verifications.add(Triplet.with("Maven", new MavenVerifier(), new FileExistsConditionChecker("pom.xml")));
         verifications.add(Triplet.with("JenkinsProjectUsers", new JenkinsProjectUserVerifier(), null));
         verifications.add(Triplet.with("Jelly", new JellyVerifier(), null));
+        verifications.add(Triplet.with("RequiredFiles", new RequiredFilesVerifier(), null));
 
         final HostingRequest hostingRequest = HostingRequestParser.retrieveAndParse(issueID);
 
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
@@ -9,14 +9,19 @@ import java.io.StringReader;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.model.DependencyManagement;
 import org.apache.maven.model.License;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.Parent;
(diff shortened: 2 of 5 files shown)


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
```diff
@@ -27,7 +27,7 @@ public class HostingChecker {
 
     public static final String INVALID_FORK_FROM = "Repository URL '%s' is not a valid GitHub repository (check that you do not have .git at the end, GitHub API doesn't support this).";
 
-    public static final Version LOWEST_JENKINS_VERSION = new Version(2, 440, 3);
+    public static final Version LOWEST_JENKINS_VERSION = new Version(2, 452, 4);
```
mawinter69, line 30: Should we go here to 2.462.3 or even 2.479.3?

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
```diff
@@ -295,6 +310,59 @@ private void checkSoftwareConfigurationManagementField(Model model, HashSet<Veri
             }
         }
     }
+
+    private void checkDependencyManagement(Model model, HashSet<VerificationMessage> hostingIssues) {
```
mawinter69, line 314: The dependencyManagement check is not 100% accurate. Ideally we would read the pom of the bom and it's transitive poms and fetch all dependencies from there to see what is managed by the bom. And then we can loop over all the dependencies and check if they are contained in the bom.
Right now it could happen that it complains about a missing dependencyManagement section although there is no dependency that could be managed via the bom.

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
```diff
@@ -295,6 +310,59 @@ private void checkSoftwareConfigurationManagementField(Model model, HashSet<Veri
...
+        JenkinsVersion jenkinsVersion = getJenkinsVersion(model);
+        if (bom.isPresent() && jenkinsVersion != null) {
+            Dependency b = bom.get();
+            String artifactId = b.getArtifactId();
+            artifactId  = artifactId.replace("${jenkins.baseline}", jenkinsVersion.baseline());
+            String expected = "bom-" + jenkinsVersion.baseline() + ".x";
+            if (!expected.equals(artifactId)) {
+                hostingIssues.add(new VerificationMessage(VerificationMessage.Severity.REQUIRED, "The artifactId `%s` of the bom is not in sync with the jenkins baseline version `%s`", artifactId, jenkinsVersion.baseline()));
+            }
+            if (!model.getProperties().containsKey("jenkins.baseline")) {
```
mawinter69, line 350: Should we enforce the usage of `jenkins.baseline` even when no dependencyManagement is required?
The archetype for an empty plugin also defines a dependeManagement section and set jenkins.baseline

timja: yes I think so

timja (approved):
LGTM

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
mawinter69: We should then also enforce the presence of the dependencyManagement I think

timja: why does dependency management need to be there if the plugin has no dependencies?

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
alecharp: Based on https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/#changing-the-minimum-required-version: `2.479.3`. 

I think there is a automatic update of that value when the documentation is built. Maybe we should use that.

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
mawinter69: The `jenkins.baseline` is also only really needed when you need the bom. And the archetype for an empty plugin also includes the bom and has no dependencies.
But I'm also fine with only enforcing the `jenkins.baseline` property

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/HostingChecker.java
mawinter69: That page states:
```
At the moment, the Jenkins releases 2.462.3 and 2.479.3 make good core dependencies. You could also consider 2.492.1 if there are specific reasons, like new features, to want something newer.
```
But the transition to Java 17 I think might justify to go to 2.479.3 now, given that many existing plugins are also updated to use 2.479.x

timja: I think so lets go with 2.479.3

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
timja: understood I think fine to leave out though, someone can start from archetype then remove bom if unneeded I don't think it should be a blocker

mawinter69: Ok so enforcing the jenkins.baseline now but not the dependencymanagement

timja (approved):

PR #4379, opened 2025-03-24
add check that there is a configuration file for dependabot, renovate or updatecli

mawinter69:


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/RequiredFilesVerifier.java
```diff
@@ -93,6 +94,14 @@ private void checkSecurityScan(GHRepository repo, HashSet<VerificationMessage> h
         }
     }
 
+    private void checkDependabot(GHRepository repo, HashSet<VerificationMessage> hostingIssues) throws IOException {
+        if (fileNotExistsInRepo(repo, ".github/dependabot.yml") &&
```
timja, line 98: I think this should also check to see if renovate is there too and be silent if its not there.

src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/RequiredFilesVerifier.java
```diff
@@ -35,6 +35,7 @@ public void verify(HostingRequest request, HashSet<VerificationMessage> hostingI
                 checkSecurityScan(repo, hostingIssues);
                 checkCodeOwners(repo, hostingIssues, forkTo);
                 checkGitignore(repo, hostingIssues);
+                checkDependabot(repo, hostingIssues);
```
alecharp, line 38: How about `doCheckDependencyBot` so it could check for either dependabot or renavate or updatecli if it is used in the future?

timja (approved):

PR #4380, opened 2025-03-24
find dependencies managed via bom

mawinter69:
Read the latest bom for the given jenkins version and extract all managed dependencies transitively. Then check the plugin dependencies and report those that are managed and have an explicit version defined.

Check that the bom version is the latest available bom version.

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
@@ -1,5 +1,6 @@
 package io.jenkins.infra.repository_permissions_updater.hosting;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.BufferedReader;
 import java.io.FileInputStream;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
```diff
@@ -312,40 +319,52 @@ private void checkSoftwareConfigurationManagementField(Model model, HashSet<Veri
...
+            Set<String> managedDependencies;
+            String bomArtifactId = "bom-" + jenkinsVersion.baseline() + ".x";
+            String latestReleasedBom = getLatestBomVersion(bomArtifactId);
+            if (latestReleasedBom  != null) {
+                managedDependencies = getManagedDependenciesFromBom(bomArtifactId, latestReleasedBom);
+            } else {
+                managedDependencies = Collections.emptySet();
+            }
+            List<Dependency> pluginDependencies = model.getDependencies().stream().filter(d -> managedDependencies.contains(getDependencyAsString(d, false))).toList();
+            if (!pluginDependencies.isEmpty() && bom.isEmpty()) {
```
timja, line 342: how will this be triggered? as bom is empty, plugin dependencies will never match the above filter?

mawinter69: The managedDependencies are collected also when no dependencyManagement section is present or it is there but doesn't include the bom. It is purely based on the jenkins version. See lines 334-340

timja (approved):

PR #4510, opened 2025-06-25
Refactor ArtifactoryPermissionsUpdater to Java edition and update access modifiers

TheMeinerLP:
This script converts one of two groovy scripts. The rework for code maintaince and readability would be done in a separate PR.
The code logic is the same as the original script.

**PR is waiting to get a review from Security Team**


src/main/java/io/jenkins/infra/repository_permissions_updater/run/PermissionJson.java
```diff
@@ -0,0 +1,17 @@
...
+        String includesPattern,
+        String excludesPattern,
+        List<String> repositories,
+        Principals principals
+) {
+    public record Principals(
+            Map<String, List<String>> users,
+            Map<String, List<String>> groups
+    ) {}
+}
```
dduportal, line 17:
```suggestion
}

```

nit: posix text file should always end with an empty endline. Ref. https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_392
Worth setting up your IDE to automatically add it when saving

Issue #4608, opened 2025-08-30
New Hosting Request: Gradle Convention Plugin for Developing Jenkins Plugins


aaravmahajanofficial:
### Repository URL

https://github.com/aaravmahajanofficial/jenkins-gradle-convention-plugin

### New Repository Name

gradle-convention-plugin

### Description

The Jenkins Gradle Convention Plugin is a Kotlin-first, 🐘 [Gradle](https://github.com/gradle/gradle) convention plugin that acts as the Maven Parent POM equivalent for Jenkins plugin development with Gradle. It provides a unified, opinionated foundation for building, testing, and publishing Jenkins plugins—standardizing best practices, automating quality checks, and eliminating boilerplate.

Built on top of the well-established [gradle-jpi-plugin](https://github.com/jenkinsci/gradle-jpi-plugin), this plugin extends JPI with extra conventions, integrated quality tools, and CI-friendly defaults—so you can focus on your plugin’s logic, not your build scripts.

✓ Stop copy-pasting boilerplate:
Get reproducible, high-quality, CI-ready Jenkins plugins out of the box!

### GitHub users to have commit permission

@aaravmahajanofficial

### Jenkins project users to have release permission

aaravmahajanofficial

### Issue tracker

GitHub issues

jenkins-cert-app:
<b>Security audit, information and commands</b>

The security team is auditing all the hosting requests, to ensure a better security by default.

This message informs you that a [Jenkins Security Scan](https://www.jenkins.io/redirect/jenkins-security-scan/) was triggered on your repository.
It takes ~10 minutes to complete.

<details><summary>Commands</summary>

The bot will parse all comments, and it will check if any line start with a command.

Security team only:
<ul>
	<li><code>/audit-ok</code> => the audit is complete, the hosting can continue :tada:.</li>
	<li><code>/audit-skip</code> => the audit is not necessary, the hosting can continue :tada:.</li>
	<li><code>/audit-findings</code> => the audit reveals some issues that require corrections :pencil2:.</li>
</ul>

Anyone:
<ul>
	<li><code>/request-security-scan</code> => the findings from the <a href="https://www.jenkins.io/redirect/jenkins-security-scan/" rel="nofollow">Jenkins Security Scan</a> were corrected, this command will re-scan your repository :mag:.</li>
	<li><code>/audit-review</code> => the findings from the audit were corrected, this command will ping the security team to review the findings :eyes:.
[...]

PR #4609, opened 2025-08-31
Create repo file for #4608 manually

NotMyFault:
Fixes https://github.com/jenkins-infra/repository-permissions-updater/issues/4608; there's something wrong in the underlying logic for non-maven distributions, but that shouldn't stop the hosting process and needs to be looked into at another point.


permissions/component-gradle-convention-plugin.yaml
```diff
@@ -0,0 +1,9 @@
+---
+name: "jenkins-gradle-convention-plugin"
```
NotMyFault, line 2:
```suggestion
name: "jenkins-gradle-convention"
```

PR #4623, opened 2025-09-06
Convert Groovy classes to Java

### LINKED ISSUE #3028: Convert groovy classes to Java
There are [a few](https://github.com/jenkins-infra/repository-permissions-updater/tree/master/src/main/groovy/io/jenkins/infra/repository_permissions_updater) Groovy classes which should be moved to Java. Not only does that align with the tooling structure, it also allows us to drop the (very dated) groovy dependencies, to prepare for a more robust future in Java.

basil:
Fixes #3028. While trying to work on this repository, I found it quite hard to follow the Groovy code. I found an old ticket for this pending migration and decided to tackle it with the help of AI. I think the result should be a lot easier to maintain than the old Groovy code.

The resulting Java is as literal a translation as is possible, designed for maximum ease in reviewing rather than any elegance in the final product. I even went out of my way to preserve small details, like 4-space indentation in the JSON, to minimize behavioral change and risk of regression. Once this PR has been merged, much further clean up is possible, but I intentionally 
kept these changes as minimal as possible. To aid reviewing, here is a diff that follows the changed files: https://basilcrow.com/tmp/diff.txt

To test this, I ensured that the generated XML files and log statements were the same before and after this PR, modulo some small things like formatting of empty lists, or ordering of developer user IDs (the originals were unordered). One area that has not been actually tested is the changes to `ArtifactoryAPI` that aren't exercised in dry run mode. I only tested in dry run mode, so I haven't tested the portions that do destructive operations. This change should not be merged until an Artifactory administrator who is familiar with our permissions model (@daniel-beck?) tests that portion, ideally by stepping through it in a debugger (without dry-run mode) and verifying that everything makes sense before any destructive operations are done.


basil:
> Beyond ensuring that my minimal data sample looked correct, I've not tested this further. IIRC, comparison generated JSON files "before" and "after" is the way to go for everything but deletion and GH repo secret generation, given how this tool is structured, and it looks like this has been done (XML being a typo?).

Thanks @daniel-beck! Yes, XML was a typo for JSON. If a member of @jenkins-infra/hosting could test GitHub repository secret generation and merge this PR, that would be great!

timja:
> If a member of @jenkins-infra/hosting could test GitHub repository secret generation and merge this PR

You should have all the access needed to test this?

From a quick look it would require creating a file with a repo in it and probably changing the main method to just do that part of the process.

basil:
I'm not familiar enough with this repository or the permissions model to test this change with confidence. I would prefer if a maintainer would handle this.

daniel-beck:
I also tested it 👍 It was just not obvious from my poorly structured comment. Expand details and I mention it there.

PR #4631, opened 2025-09-10
Don't retry creating secrets on first run

timja:
Noticed when testing https://github.com/jenkins-infra/repository-permissions-updater/pull/4623

On the first time creating the token it will return 201 and not 204.

Only affects the first run of a repo after CD is enabled.

Tested locally, output seems fine. (Haven't managed to trigger a retry though)


mawinter69 (approved):

NotMyFault (approved):
Looks good

PR #4752, opened 2025-12-02
Clean up CloudBees permissions for jenkinsci/ant-plugin CAP removal

mikecirioli:
This PR cleans up CloudBees-specific permissions for **jenkinsci/ant-plugin** following its removal from the CloudBees Assurance Program (CAP).

### DIFF
--- a/permissions/plugin-ant.yml
+++ b/permissions/plugin-ant.yml
@@ -12,6 +12,3 @@ developers:
   - "jglick"
   - "fcojfernandez"
   - "markewaite"
-security:
-  contacts:
-    jira: []


permissions/plugin-ant.yml
```diff
@@ -10,8 +10,7 @@ cd:
 developers:
   - "armfergom"
   - "jglick"
-  - "fcojfernandez"
```
alecharp: Is this related?

mikecirioli: It was suggested by the automation tooling i am using, but i am currently trying to find out why.  I suspect it was intentional - i will confirm.

timja:
> I think I saw a comment from Tim [recently](https://github.com/jenkins-infra/repository-permissions-updater/pull/4695#discussion_r2481113957) that the whole section should be removed, not just blank-ed.
> 
> Otherwise 👍

yes that's right, the section shouldn't be there if its not used.

PR #4765, opened 2025-12-04
Migrate gerrit trigger plugin issues to GitHub

timja:
Hi

@rsandell for approval


rsandell (approved):
I'm slightly worried about more GitHub spam though, as if I didn't have alot of that already 😅

PR #4768, opened 2025-12-04
Migrate active directory plugin issues to GitHub

timja:
Hi

@batmat / @jtnord / @rsandell  for approval

### DIFF
--- a/permissions/plugin-active-directory.yml
+++ b/permissions/plugin-active-directory.yml
@@ -1,8 +1,8 @@
 ---
 name: "active-directory"
-github: "jenkinsci/active-directory-plugin"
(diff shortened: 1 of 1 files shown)


timja:
> The Jenkins thread was about migrating core to GH issues, wasn't it? AIUI, this would aim at kinda taking issues.jenkins.io down eventually?

Aye, its called out in the JEP:
https://github.com/jenkinsci/jep/blob/master/jep/238/README.adoc#project-management-in-one-view

> It is expected that all repositories will transition from Jira to GitHub issues after the core components migration. This will alleviate the temporary issues as all issues will now be in one place.

PR #4783, opened 2025-12-07
Migrate Active Choices plugin issues to GitHub

timja:
Hi

@kinow  for approval

### DIFF
--- a/permissions/plugin-uno-choice.yml
+++ b/permissions/plugin-uno-choice.yml
@@ -1,8 +1,8 @@
 ---
 name: "uno-choice"
-github: "jenkinsci/active-choices-plugin"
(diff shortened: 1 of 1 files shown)


timja:
> BTW, will it magically create issues for the JIRA issues that we already have? And maybe have the original JIRA number in the issue somewhere? I keep a folder in my email inbox with the issues I need to work for next releases, and it'd be useful if I could easily find them on GH.
> 
> Thanks again.

Yes see https://github.com/jenkinsci/jenkins/issues for what they look like.

The easiest way to map old to new is to use the redirect service:
https://issue-redirect.jenkins.io/issue/1

or just drop the Jira issue key into the github issue search for the repo

kinow:
Thank you, @imoutsatsos !

PR #4784, opened 2025-12-07
Migrate xcode-plugin plugin issues to GitHub

timja:
Hi

@kazuhidet for approval


flabrie:
Hi @timja,

Do you think it would it be possible to migrate Jira to GitHub, now?

PR #4794, opened 2025-12-09
Migrate warnings-ng plugin issues to GitHub

timja:
Hi

@uhafner for approval


uhafner:
Sorry, my comment should be here:

Can we also activate this migration process for the analysis-model (the model for the warnings plugin)? So migrate analysis-model Jira to analysis-model GitHub?
These issues are typically linked with the warnings plugin, so having only one component in GitHub makes no sense.

PR #4795, opened 2025-12-09
Migrate warnings plugin issues to GitHub

timja:
Hi

@uhafner what do you think? I was thinking of unarchiving, migrating and rearchiving the repo, just so all the issue are on GitHub?

### DIFF
--- a/permissions/plugin-warnings.yml
+++ b/permissions/plugin-warnings.yml
@@ -1,8 +1,8 @@
 ---
 name: "warnings"
-github: "jenkinsci/warnings-plugin"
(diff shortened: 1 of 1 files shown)


uhafner (approved):
Would it be possible to move the old warnings issues to jenkinsci/warnings-ng-plugin as well? Or will the script overwrite all existing issues?

uhafner:
Can we also activate this migration process for the analysis-model (the model for the warnings plugin)? So migrate analysis-model Jira to analysis-model GitHub? 
These issues are typically linked with the warnings plugin, so having only one component in GitHub makes no sense.

PR #4803, opened 2025-12-09
Migrate jobConfigHistory plugin issues to GitHub

timja:
Hi

@StefanSpieker  for approval

### DIFF
--- a/permissions/plugin-jobConfigHistory.yml
+++ b/permissions/plugin-jobConfigHistory.yml
@@ -1,8 +1,8 @@
 ---
 name: "jobConfigHistory"
-github: "jenkinsci/job-config-history-plugin"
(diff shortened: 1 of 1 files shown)


StefanSpieker (approved):

timja:
> Hi @timja - that is a great idea, anything I can do to support you?

Hey should be all good.

I've AI'ed a new script to orchestrate running these migrations and just need to test it so that I can move from me running a bunch of scripts per repo to providing a list of repos and having it all done for me.

If that works well then this will be much simpler.

timja:
I now have a script that fully automates with no manual steps: https://github.com/lemeurherve/jira-issues-importer/pull/76

now its just a matter of creating text files with plugins to migrate in batches.

PR #4868, opened 2025-12-31
Run on java 25

NotMyFault:
As the title says.

I've removed the `withEnv` clause due to https://github.com/jenkins-infra/jenkins-infra/blob/f67283f8dc6dac49954ba5407796bce2e1562eb8/hieradata/clients/controller.trusted.ci.jenkins.io.yaml#L176


NotMyFault:
Confirmed on trusted.ci that builds continue working as expected.

PR #4896, opened 2026-01-14
check for ban of commons-lang-2

mawinter69:
Ensure that the commons-lang-2 ban check is included in the pom and not set to skipped.
So we ensure that new plugins are not using it and we can proceed with it's removal in core via https://github.com/jenkinsci/jenkins/pull/26105

### DIFF
--- a/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
+++ b/src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
@@ -526,7 +526,7 @@ public class MavenVerifier implements BuildSystemVerifier {
                 || !props.getProperty("ban-commons-lang-2.skip").equals("false")) {
             hostingIssues.add(new VerificationMessage(
                     VerificationMessage.Severity.REQUIRED,
-                    "Please define the property `ban-commons-lang-2.skips` and set it to `false`. This should help prevent accidental usage of the deprecated commons-lang-2 library that is "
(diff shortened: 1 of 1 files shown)


src/main/java/io/jenkins/infra/repository_permissions_updater/hosting/MavenVerifier.java
```diff
@@ -522,6 +522,13 @@ private void checkProperties(Model model, HashSet<VerificationMessage> hostingIs
                             "Please define the property `hpi.strictBundledArtifacts` and set it to `true`. This should help prevent accidental library bundling when adding and updating dependencies."
                                     + "See [Bundling third-party libraries](https://www.jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/#bundling-third-party-libraries)."));
         }
+        if (!props.containsKey("ban-commons-lang-2.skip")
+                || !props.getProperty("ban-commons-lang-2.skip").equals("false")) {
+            hostingIssues.add(new VerificationMessage(
+                    VerificationMessage.Severity.REQUIRED,
+                    "Please define the property `ban-commons-lang-2.skips` and set it to `false`. This should help prevent accidental usage of the deprecated commons-lang-2 library that is "
```
timja:
```suggestion
                    "Please define the property `ban-commons-lang-2.skip` and set it to `false`. This should help prevent accidental usage of the deprecated commons-lang-2 library that is "
```

alecharp (approved):
Good idea.

PR #5159, opened 2026-07-16
Enable CD for testingbot plugin

jochen-testingbot:
Enables JEP-229 continuous delivery for the `testingbot` plugin (`jenkinsci/testingbot-plugin`).

The plugin has been set up for CD on the repository side (incrementals versioning via `.mvn/extensions.xml` + `.mvn/maven.config`, `<version>${revision}.${changelist}</version>`, and `.github/workflows/cd.yaml` from the standard template). This PR adds `cd: enabled: true` so the `MAVEN_USERNAME`/`MAVEN_TOKEN` secrets are provisioned and automated releases can run.

Repository CD setup PR: https://github.com/jenkinsci/testingbot-plugin (modernization merged to master; cd.yaml + incrementals in place).

### DIFF
--- a/permissions/plugin-testingbot.yml
+++ b/permissions/plugin-testingbot.yml
@@ -5,6 +5,8 @@ issues:
   - jira: '17570'  # testingbot-plugin
 paths:
   - "testingbot/testingbot"
+cd:
+  enabled: true
 developers:
   - "testingbot"
   - "jochen_testingbot"


timja:
Tag should be `scmHead`: https://github.com/jenkinsci/archetypes/blob/48fa8b1d5983c90bca56eb6e112f89d40f5a367c/empty-plugin/src/main/resources/archetype-resources/pom.xml#L29

cd.yaml is missing the username secret:
https://github.com/jenkinsci/archetypes/blob/master/common-files/.github/workflows/cd.yaml

PR #5191, opened 2026-08-02
Add a hosting security scan action as replacement for the bot

### LINKED ISSUE #5189: Run security scan from hosting repo in GitHub actions

Hi @jenkins-infra/security is there anything stopping us from getting the security scan moved to Jenkins-infra / this repo and running it directly during hosting?

It’s a pain when it’s breaks using the existing setup and no logging available directly isn’t great either

### Upstream changes

_No response_

mawinter69:
Mostly done by claude

fixes #5189

Runs on hosting issue creation and on the usual comment.

Currently the mapping from ruleId to the docs and the title is hard coded in the script. Probably would be better to have that mapping in the jenkins codeql repo and fetch that

The initial message is fetched from the repo (github didn't like it to be inline)

When merging the existing trigger would need to be disabled.

On my tests what was annoying is that it prints many warnings of the form:
```
This run of the CodeQL Action does not have permission to access the CodeQL Action API endpoints. This could be because the Action is running on a pull request from a fork. If not, please ensure the workflow has at least the 'security-events: read' permission. Details: Resource not accessible by integration - https://docs.github.com/rest
```
And also an error
```
Maximum retry attempts exhausted (4), aborting database upload
```
though the error doesn't stop the action from running fine.

Sample: https://github.com/mawinter69/rpu2/issues/12

### DIFF
--- a/.github/workflows/hosting-security-scan.yaml
+++ b/.github/workflows/hosting-security-scan.yaml
@@ -62,6 +62,10 @@ jobs:
         with:
           repository: ${{ steps.extract-repo.outputs.repo }}
 
+      - name: Capture plugin repository commit SHA
+        id: plugin-sha
+        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
       - name: Set up Java
         uses: actions/setup-java@v5.6.0
         with:
@@ -125,7 +129,7 @@ jobs:
             const DOC_BASE = 'https://github.com/jenkins-infra/jenkins-codeql/blob/main/src/';
 
             const sarif = JSON.parse(fs.readFileSync('jenkins-security-scan.sarif', 'utf8'));
-            const repoRef = sarif.runs?.[0]?.versionControlProvenance?.[0]?.revisionId || 'HEAD';
+            const repoRef = sarif.runs?.[0]?.versionControlProvenance?.[0]?.revisionId || '${{ steps.plugin-sha.outputs.sha }}';
 
             // Group results by ruleId
             const byRule = new Map();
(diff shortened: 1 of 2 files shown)


.github/workflows/hosting-security-scan.yaml
```diff
@@ -0,0 +1,197 @@
...
+              for (const [ruleId, findings] of byRule) {
+                const check = CHECKS[ruleId];
+                const title = check?.title || ruleId;
+                const docUrl = check ? DOC_BASE + check.doc : null;
+                const docLine = docUrl
+                  ? `You can find detailed information about this finding [here](${docUrl}).`
+                  : '';
+                const findingItems = findings.map(f => {
+                  const summary = f.githubLink ? `<a href="${f.githubLink}">${f.linkLabel}</a>` : f.linkLabel;
+                  return `<details><summary>${summary}</summary>\n\n\`\`\`\n${f.message}\n\`\`\`\n\n</details>`;
```
timja: (not a blocker)

instead of wrapping in a code block might be nice to quote it or just output directly?

Its a bit hard to read right now:

Image

timja (approved):
Looks great, whats outstanding on this?

.github/workflows/hosting-security-scan.yaml
mawinter69: That hasn't changed. It was like that also before, but I changed it to use single backticks, that will make wrapping

mawinter69:
> Looks great, whats outstanding on this?

The links used `HEAD` instead of the git hash

timja (approved):

daniel-beck:
> Currently the mapping from ruleId to the docs and the title is hard coded in the script. Probably would be better to have that mapping in the jenkins codeql repo and fetch that

Mapping rule ID to doc is something the current script also needs to do, but the title is just `.shortDescription.text` in the current iteration.

Issue #5243, opened 2026-08-28
Hosting request for github-pr-approval-plugin


olamy:
### Repository URL

https://github.com/olamy/github-pr-approval-plugin

### New Repository Name

github-pr-approval-plugin

### Description

  Adds a new fork pull request trust policy — *External approval required* — to the
  GitHub Branch Source plugin's "Discover pull requests from forks" trait. A fork PR is
  discovered as usual, but its branch job stays disabled with a *Pending Approval* marker
  until someone with Configure permission approves it, at which point the job is enabled
  and a build is scheduled. Optional settings allow re-approval on each new commit, and
  auto-approval by GitHub login or PR label.

  The plugin depends on a released GitHub Branch Source plugin and contributes the policy
  purely as an `@Extension` (an `SCMHeadAuthority`), so it needs no change to GitHub Branch
  Source. It resolves the review discussion in jenkinsci/github-branch-source-plugin#1556,
  where a standalone plugin was suggested rather than pulling `branch-api` into the branch
  source plugin.

### GitHub users to have commit permission

@olamy 

### Jenkins project users to have release permission

olamy

### Automated release via GitHub Actions (recommended)

Yes

jenkins-cert-app:
<b>Security audit, information and commands</b>

The security team is auditing all the hosting requests, to ensure a better security by default.

This message informs you that a [Jenkins Security Scan](https://www.jenkins.io/redirect/jenkins-security-scan/) was triggered on your repository.
It takes ~10 minutes to complete.

<details><summary>Commands</summary>

The bot will parse all comments, and it will check if any line start with a command.

Security team only:
<ul>
	<li><code>/audit-ok</code> => the audit is complete, the hosting can continue :tada:.</li>
	<li><code>/audit-skip</code> => the audit is not necessary, the hosting can continue :tada:.</li>
	<li><code>/audit-findings</code> => the audit reveals some issues that require corrections :pencil2:.</li>
</ul>

Anyone:
<ul>
	<li><code>/request-security-scan</code> => the findings from the <a href="https://www.jenkins.io/redirect/jenkins-security-scan/" rel="nofollow">Jenkins Security Scan</a> were corrected, this command will re-scan your repository :mag:.</li>
	<li><code>/audit-review</code> => the findings from the audit were corrected, this command will ping the security team to review the findings :eyes:.
	It's only applicable when the previous audit required changes.</li>
</ul>

<i>Only one command can be requested per comment.</i>

[...]

PR #5244, opened 2026-08-28
Add release permissions for github-pr-approval plugin

olamy:
Adds release permissions (CD enabled) for the `io.jenkins.plugins:github-pr-approval` plugin.

Follows hosting request #5243.


mawinter69:
The hosting process takes care to create the file automatically when hosting is accepted

